PR #10, opened 2022-02-25
Make namespace idempotent

TerryHowe:


lewisdiamond:
/lgtm
Noting that the following would reset the namespace to the default every time if that's something we need in the future.
`kubectl create namespace eksa-packages --dry-run=client -o yaml | kubectl apply -f -`
`kubectl create namespace eksa-packages --dry-run=client -o yaml | kubectl delete -f -`

lewisdiamond (approved):

PR #11, opened 2022-02-25
Use version in package status

TerryHowe:
Set Status.Version to human readable version

```
NAME      PACKAGE   AGE     STATE       VERSION   DETAIL
my-test   Test      5m47s   installed   v0.1.0 
```

### DIFF
--- a/controllers/package_controller.go
+++ b/controllers/package_controller.go
@@ -128,7 +128,7 @@ func (r *PackageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			return ctrl.Result{RequeueAfter: retryLong}, err
 		}
 
-		managerContext.Source, err = bundle.FindSource(managerContext.Package.Spec.PackageName, managerContext.Package.Spec.PackageVersion)
+		managerContext.Source, managerContext.Version, err = bundle.FindSource(managerContext.Package.Spec.PackageName, managerContext.Package.Spec.PackageVersion)
 		if err != nil {
 			return ctrl.Result{RequeueAfter: retryLong}, err
 		}
--- a/controllers/packagebundlecontroller_controller.go
+++ b/controllers/packagebundlecontroller_controller.go
@@ -123,8 +123,7 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 
 	latestBundle, err := r.bundleManager.LatestBundle(ctx, abc.Spec.Source.BaseRef())
 	if err != nil {
-		r.Log.Error(err, "Error getting latest bundle")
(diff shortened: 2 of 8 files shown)


controllers/packagebundlecontroller_controller.go
```diff
@@ -123,7 +123,8 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 
 	latestBundle, err := r.bundleManager.LatestBundle(ctx, abc.Spec.Source.BaseRef())
 	if err != nil {
-		return result, err
+		r.Log.Error(err, "Error getting latest bundle")
```
ewollesen, line 126: Why this change? Was the error not being logged previously?

ewollesen: I can merge regardless, but I'm curious.

controllers/package_controller.go
```diff
@@ -128,7 +128,7 @@ func (r *PackageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			return ctrl.Result{RequeueAfter: retryLong}, err
 		}
 
-		managerContext.Source, err = bundle.FindSource(managerContext.Package.Spec.PackageName, managerContext.Package.Spec.PackageVersion)
+		managerContext.Source, managerContext.Version, err = bundle.FindSource(managerContext.Package.Spec.PackageName, managerContext.Package.Spec.PackageVersion)
```
lewisdiamond, line 131: I think you can avoid passing down the `version.Name` from `FindSource` all the way to here by just using `managerContext.Package.Spec.PackageVersion` if a source was found.

lewisdiamond: It's my opinion that if the version installed was specified as a `sha256:tag`, we should display that. In the current implementation, it would use the `version.Name`, but if you use `managerContext.Package.Spec.PackageVersion`, it would display the `sha256:tag` if that's what was specified (those cases are going to be rare of course).

controllers/packagebundlecontroller_controller.go
TerryHowe: I didn't mean to include that change.

controllers/package_controller.go
TerryHowe: The Package.Spec.PackageVersion is the requested version, not necessarily the installed version. The sha is in managerContext.Source

TerryHowe: I added version here from the bundle `name` because it is the human readable version. Otherwise the user would just have a sha and no idea what that releates to without looking it up in the bundle

ewollesen (approved):

PR #13, opened 2022-02-28
Upgrades to fix a dependabot issue.

ewollesen:
https://github.com/aws/eks-anywhere-packages/security/dependabot/1

Is caused by our version of github.com/docker/distribution, which is a
dependency of oras.land/oras-go. Upgrading oras required some changes to
the code, as well as our version of the helm.sh/helm/v3. That in turn
requires upgrades to a bunch of packages, including
sigs.k8s.io/controller-runtime and github.com/go-logr/logr. Those
changes required more tweaks to the code.

*Issue #, if available:*

*Description of changes:*


TerryHowe (approved):
/lgtm

controllers/package_controller_test.go
```diff
@@ -6,11 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-logr/logr"
```
a-cool-train, line 9: Why are we switching out loggers?

ewollesen: > Why are we switching out loggers?

We aren't actually switching out loggers. One of the packages upgraded was the logr package. In the new version, they've moved `logr.Logger` from an interface to a struct. As a result, we can't pass around `nil` when we don't care about a logger. As a result, I can't use `nil` in these tests anymore, I have to use `logr.Discard()` instead. So no we're having to import that package in these tests, where before we didn't.

PR #18, opened 2022-03-01
Adding Helm chart for controller

jonahjon:

*Issue #, if available:*

*Description of changes:*

Helm Chart!!!


a-cool-train:
is this binary: charts/eks-anywhere-packages-0.0.0.tgz required?

charts/eks-anywhere-packages/templates/NOTES.txt
```diff
@@ -0,0 +1 @@
+1. These are notes:
```
a-cool-train, line 1: Is this complete?

TerryHowe:
> is this binary: charts/eks-anywhere-packages-0.0.0.tgz required?

Yeh, remove that and maybe add tgz to the gitignore

charts/eks-anywhere-packages/README.md
```diff
@@ -0,0 +1,65 @@
...
+```bash
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+```
+
+To do a simple install run:
+
+```bash
+helm upgrade --install --namespace eksa-packages --create-namespace \
+  eksa-packages eks-anywhere-packages/ \
+  --values values.yaml
```
TerryHowe, line 20: I would expect the chart to create and use the namespace and the namespace to be overridden possibly in values.yaml

charts/eks-anywhere-packages/README.md.gotmpl
```diff
@@ -0,0 +1,30 @@
+{{ template "chart.header" . }}
+{{ template "chart.description" . }}
+
+{{ template "chart.versionBadge" . }}{{ template "chart.typeBadge" . }}{{ template "chart.appVersionBadge" . }}
```
TerryHowe, line 4: ah, maybe don't check in the README.md

charts/eks-anywhere-packages/templates/_helpers.tpl
```diff
@@ -0,0 +1,77 @@
...
+{{/*
+Define the eks-anywhere-packages.namespace template if set with namespace or .Release.Namespace is set
+*/}}
+{{- define "eks-anywhere-packages.namespace" -}}
+{{- if .Values.namespace -}}
+{{ printf "namespace: %s" .Values.namespace }}
+{{- else -}}
+{{ printf "namespace: %s" .Release.Namespace }}
+{{- end -}}
+{{- end -}}
```
TerryHowe, line 77: Maybe add something like https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/aws/eks-anywhere-test/eks-anywhere-test/templates/_helpers.tpl#L63:L73

charts/eks-anywhere-packages/templates/certs.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: cert-manager.io/v1
```
TerryHowe, line 1: These files are just copies, right? Might be nice if there was a build step to copy them in rather than duplicate

charts/eks-anywhere-packages/templates/deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+          - /manager
+          args:
+          - --health-probe-bind-address=:8081
+          - --metrics-bind-address=127.0.0.1:8080
+          - --leader-elect
+          {{- with .Values.controller.securityContext }}
+          securityContext:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          image: {{ .Values.controller.image }}
```
TerryHowe, line 50: Use the helper to generate the image name here maybe

TerryHowe: https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/aws/eks-anywhere-test/eks-anywhere-test/templates/deployment.yaml#L24

charts/index.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
TerryHowe, line 1: This looks generated, maybe not checked in?

charts/eks-anywhere-packages/values.yaml
```diff
@@ -0,0 +1,89 @@
+# Default values for eks-anywhere-packages.
+# This is a YAML-formatted file.
```
TerryHowe, line 2: We might want to be more restrictive on what we allow to set here.

hack/helm.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
```
TerryHowe, line 2: I was expecting a mark target at the top level `make chart` or `make helm` something like that.

charts/eks-anywhere-packages/README.md
```diff
@@ -0,0 +1,65 @@
...
+A Helm chart for eks-anywhere-packages, a controller for managing the lifecycle of EKS Anywhere packages.
+
+![Version: 0.0.0](https://img.shields.io/badge/Version-0.0.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.0.0](https://img.shields.io/badge/AppVersion-0.0.0-informational?style=flat-square)
+
+## Installing the Chart
+
+Before the chart can be installed the we need to install cert-manager.
+
+```bash
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
```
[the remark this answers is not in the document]
jonahjon: I think the helm install would be an ok alternative, but does it require AWS ECR auth for this specific command?

charts/eks-anywhere-packages/README.md
jonahjon: You can override the namespace in the values actually, but helm isn't good at creating the namespace via a template so most people do it this way, or create the namespace beforehand with a command.

charts/eks-anywhere-packages/templates/certs.yaml
jonahjon: Yea I think that can come after, but they are just copies with templating for the namespace, and annotations.

charts/eks-anywhere-packages/values.yaml
TerryHowe: That was a general comment on the values.yaml. Especially for our first release, we want to have as few moving parts as possible and hard code some of these things. This will need the ability to override registry for the image and tag/digest of our image though

charts/eks-anywhere-packages/values.yaml
```diff
@@ -0,0 +1,89 @@
...
+    initialDelaySeconds: 15
+    periodSeconds: 20
+  # -- readinessProbe controls the values for controller container readiness. TODO Add in templating later
+  readinessProbe:
+    path: /readyz
+    port: 8081
+    initialDelaySeconds: 5
+    periodSeconds: 10
+  # -- Controller image.
+  image: "public.ecr.aws/l0g8r8j6/eks-anywhere-packages"
```
TerryHowe, line 65: It would be easier to break this up into registry, repository and tag/digest

charts/eks-anywhere-packages/values.yaml
```diff
@@ -0,0 +1,89 @@
...
+  resources:
+    requests:
+      cpu: 100m
+      memory: 50Mi
+    limits:
+      cpu: 500m
+      memory: 100Mi
+kuberbacproxy:
+  # -- Webhook image.
+  image: "gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0"
```
TerryHowe, line 80: registry, repository and tag/digest here as well and default out ours

charts/eks-anywhere-packages/README.md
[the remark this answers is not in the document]
TerryHowe: No, that is public ECR. You might see a failure if you have logged in and your creds are stale though

charts/eks-anywhere-packages/README.md
a-cool-train: does the order of the namespace and create-namespace flag matter? If not, we could provide create-namespace first, then namespace after to make the creation clear?

charts/eks-anywhere-packages/templates/crd.yaml
```diff
@@ -0,0 +1,488 @@
...
+    subresources:
+      status: {}
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
+---
+apiVersion: apiextensions.k8s.io/v1
```
a-cool-train, line 96: do all the crds need to be on the same file? Is it possible to split them into separate files for readability?

charts/eks-anywhere-packages/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+To Uninstall:
+
+```
+helm delete eksa-packages
+```
+
+## Values
+
+| Key | Type | Default | Description |
```
a-cool-train, line 31: Is these values section required in the README? It clutters the README without providing usability of the application in question.

jonahjon: It's pretty standard practice for helm charts, it's from an autogenerated docs tool. Once you see it in Github it looks alot better. It will look like this https://github.com/aws/karpenter/tree/main/charts/karpenter

charts/eks-anywhere-packages/templates/crd.yaml
jonahjon: It is possible to split them up, was trying to reduce the total file count, but I could also create a /crds folder potentially.

charts/eks-anywhere-packages/README.md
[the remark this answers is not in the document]
jonahjon: I put you suggestion in place for now, but once we get things moved to prod we can change it.

TerryHowe (approved):
/lgtm

charts/eks-anywhere-packages/templates/crd.yaml
a-cool-train: I think the /crds folder will make it easier to naviage the files. It is just my thought though.

PR #21, opened 2022-03-02
Updating helm driver to be compatible with helm3.8

lewisdiamond:
Helm was upgraded to 3.8 but our helm driver required changes in order to be
compatible. This commit instantiates a `registry.Client` to be placed in the
configuration object and used by Install actions.

### DIFF
--- a/pkg/driver/helmdriver.go
+++ b/pkg/driver/helmdriver.go
@@ -33,7 +33,7 @@ func NewHelm(log logr.Logger) (*helmDriver, error) {
 	//see: https://github.com/aws/eks-anywhere-packages/issues/20
 	client, err := registry.NewClient()
 	if err != nil {
-		return nil, fmt.Errorf("Error creating registry client while initializing helm driver: %w", err)
+		return nil, fmt.Errorf("creating registry client while initializing helm driver: %w", err)
 	}
 	cfg := &action.Configuration{RegistryClient: client}
 	err = cfg.Init(settings.RESTClientGetter(), settings.Namespace(),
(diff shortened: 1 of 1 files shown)


pkg/driver/helmdriver.go
```diff
@@ -28,8 +29,14 @@ var _ PackageDriver = (*helmDriver)(nil)
 
 func NewHelm(log logr.Logger) (*helmDriver, error) {
 	settings := cli.New()
-	cfg := &action.Configuration{}
-	err := cfg.Init(settings.RESTClientGetter(), settings.Namespace(),
+	//TODO: Allow configuring the client's credentials from a secret
+	//see: https://github.com/aws/eks-anywhere-packages/issues/20
+	client, err := registry.NewClient()
+	if err != nil {
+		return nil, fmt.Errorf("Error creating registry client while initializing helm driver: %w", err)
```
ewollesen, line 36: nit: There's no need to prefix this with "Error".

ewollesen (approved):
I can approve this as is, but I figured I'd let you choose if you want to deal with my nit or not.

pkg/driver/helmdriver.go
TerryHowe:
```suggestion
		return nil, fmt.Errorf("creating registry client while initializing helm driver: %w", err)
```

TerryHowe: Lower case too

TerryHowe (approved):
/lgtm

PR #34, opened 2022-03-16
Add image list to bundle

TerryHowe:


config/crd/bases/packages.eks.amazonaws.com_packagebundles.yaml
```diff
@@ -76,12 +76,33 @@ spec:
...
+                                      description: Digest is a checksum value identifying
+                                        the version of the package and its contents.
+                                      type: string
+                                    repository:
+                                      description: Repository is source in the registry
+                                        for the image
+                                      type: string
+                                  required:
+                                  - digest
+                                  - repository
```
lewisdiamond, line 96: I think requiring repository makes sense but we have to re-organize much of what we currently have since we're sharing repositories for multiple images as well as the helm chart. We need to validate how harbor and ecr deal with nested repositories and settle on a standard, i.e. `registry/harbor/helm:1.0.0` + `registry/harbor/images/harbor:1.0.0`

lewisdiamond:
We will also have to add these images to the `excludes` field of the bundle signature. My understanding is that these would not be used at all during runtime and that this is only useful for synchronizing artifacts.

config/crd/bases/packages.eks.amazonaws.com_packagebundles.yaml
TerryHowe: If we need to push helm charts to their own repositories, that is fine, but that is not a topic for this PR

jonahjon: We filter by OCI Image Media Types currently when using the same repo for bundle creation. (Helm vs Image) so this should work just fine for that.

jonahjon (approved):

ewollesen:
I'm pretty sure you've told me this before, but I don't recall: who is the consumer of this information? That is, why is this information useful to put into the bundle?

ewollesen (approved):
/lgtm

PR #38, opened 2022-03-17
Update helm crds

TerryHowe:


Makefile
```diff
@@ -97,7 +97,7 @@ docker-build: test ## Build docker image with the package-manager.
 docker-push: ## Push docker image with the package-manager.
 	docker push ${IMG}
 
-helm-build:
+helm/build:
```
TerryHowe, line 100: this is what the target is in build tooling

jonahjon (approved):

PR #40, opened 2022-03-17
Update go.mod file

TerryHowe:
Am I only the one seeing this?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/aws/eks-anywhere-packages
 go 1.17
 
 require (
-	github.com/containerd/containerd v1.6.1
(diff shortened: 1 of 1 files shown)


jonahjon:
I'm not sure what this change is really doing.

Is it just moving it to the indirect portion?

ewollesen:
> I'm not sure what this change is really doing.
> 
> Is it just moving it to the indirect portion?

Yes, that's all it is doing.

I've seen this too. It probably means someone merged a change without including an updated go.mod. It happens sometimes. That or slightly different versions of go are installed on peoples' dev machines... :thinking:

/lgtm

ewollesen (approved):

PR #43, opened 2022-03-18
Helm install with name

TerryHowe:


jonahjon:
It looks like we still keep the digest here right?

jonahjon:
Looks good but the title is a bit confusing

api/v1alpha1/packagebundle_test.go
```diff
@@ -34,6 +34,7 @@ func TestPackageBundle_Find(t *testing.T) {
 	expected := api.PackageOCISource{
 		Registry:   "public.ecr.aws/l0g8r8j6",
 		Repository: "eks-anywhere-test",
+		Name:       "v0.1.0",
```
jonahjon, line 37: remove the v in front of 0.1.0 here I think

TerryHowe: The v in the version is a problem with the helm Chart.yaml and this is just test data anyway that isn't even used

TerryHowe:
> Looks good but the title is a bit confusing

Fair enough, I changed the title to "Helm install with name"

api/v1alpha1/packagebundle.go
```diff
@@ -8,6 +8,7 @@ import (
 type PackageOCISource struct {
 	Registry   string `json:"registry"`
 	Repository string `json:"repository"`
+	Name       string `json:"name"`
```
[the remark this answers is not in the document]
TerryHowe: `Name` is from `Name` in the bundle

ewollesen:
I don't have a problem merging this, but since we just found a way to sidestep helm and it's semver stuff, I'm wondering if we need this anymore?

TerryHowe:
> I don't have a problem merging this, but since we just found a way to sidestep helm and it's semver stuff, I'm wondering if we need this anymore?


Agreed, hopefully we don't need this, but I'll leave it up in case the other approach runs into a snag.

PR #45, opened 2022-03-19
Setup envtest

TerryHowe:
https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#where-does-it-put-all-those-binaries

Build is broken for arm64 mac


Makefile
```diff
@@ -73,9 +73,8 @@ GOTESTS ?= ./...
 # go help testflags to see all options.
 GOTESTFLAGS ?= ""
 test: manifests generate vet mocks ${SIGNED_ARTIFACTS} ## Run tests.
-	mkdir -p ${ENVTEST_ASSETS_DIR}
-	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.8.3/hack/setup-envtest.sh
-	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); $(GO) test $(GOTESTFLAGS) `$(GO) list $(GOTESTS) | grep -v mocks` -coverprofile cover.out
+	go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
ewollesen, line 76: Shouldn't this be done somewhere that it can be a one-time thing? 

It should probably specify a version too, `@latest` tends to come back and bite one's hind-quarters. :)

Lastly, this should probably use `$(GO)`? Or I guess it maybe doesn't matter if it's just installing a binary to be used elsewhere.

TerryHowe: I need to address these concerns yet.

PR #46, opened 2022-03-21
Restore signature webhook config

lewisdiamond:
This config was thought not to be used, but it is at least used during
the webhook envtests.


TerryHowe (approved):
/lgtm

PR #47, opened 2022-03-21
Add EKSA_PUBLIC_KEY environment variable

lewisdiamond:
This env var allows overriding the public key used for packages releases
signature validation.

### DIFF
--- a/api/testdata/bundle_two.yaml.signed
+++ b/api/testdata/bundle_two.yaml.signed
@@ -3,7 +3,7 @@ kind: PackageBundle
 metadata:
   name: v1-21-1002
   namespace: eksa-packages
-  annotations: {}
+  annotations: {eksa.aws.com/signature: MEQCIFjkZDN7bd4hCI8pB4EvoPpJzVRMW4t7vl0ef/D/YhaiAiBb4QJXmiw1DXrvXcqIX1cWaWbWTRo7R496EWsCIL+2lg==}
 spec:
   kubeVersion: v1.21
   packages:
--- a/api/v1alpha1/packagebundle_webhook.go
+++ b/api/v1alpha1/packagebundle_webhook.go
@@ -17,12 +17,17 @@ package v1alpha1
 import (
 	"encoding/base64"
 	"errors"
+	"os"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	"github.com/aws/eks-anywhere-packages/pkg/signature"
+)
+
+const (
+	PublicKeyEnvVar = "EKSA_PUBLIC_KEY"
 )
 
 // apilog is for logging in this package.
@@ -50,10 +55,19 @@ func (r *PackageBundle) ValidateDelete() error {
 }
 
 func (r *PackageBundle) validate() error {
-	valid, digest, yml, err := signature.ValidateSignature(r, signature.EksaDomain)
-	if err == nil && !valid {
-		apilog.Info("Invalid signature", "Error", err, "Digest", base64.StdEncoding.EncodeToString(digest[:]), "Manifest", string(yml))
-		err = errors.New("The signature is invalid for the configured public key")
+	//TODO Turn this into fields in a new CRD used for signature validation configuration
+	keyOverride := os.Getenv(PublicKeyEnvVar)
+	domain := signature.EksaDomain
+	if keyOverride != "" {
+		domain = signature.Domain{Name: signature.DomainName, Pubkey: keyOverride}
 	}
-	return err
+	valid, digest, yml, err := signature.ValidateSignature(r, domain)
+	if err != nil {
+		return err
+	}
+	if !valid {
+		apilog.Info("Invalid signature", "Error", err, "Digest", base64.StdEncoding.EncodeToString(digest[:]), "Manifest", string(yml))
+		return errors.New("The signature is invalid for the configured public key: " + domain.Pubkey)
+	}
+	return nil
 }
--- a/api/v1alpha1/webhook_suite_test.go
+++ b/api/v1alpha1/webhook_suite_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package v1alpha1
+package v1alpha1_test
 
 import (
 	"context"
@@ -29,15 +29,20 @@ import (
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 
 	//+kubebuilder:scaffold:imports
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-
-	//	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	. "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+	"github.com/aws/eks-anywhere-packages/pkg/bundle"
+	"github.com/aws/eks-anywhere-packages/pkg/signature"
+	"github.com/aws/eks-anywhere-packages/pkg/testutil"
 )
 
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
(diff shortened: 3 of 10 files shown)


config/webhook/manifests.yaml
```diff
@@ -26,3 +26,23 @@ webhooks:
     resources:
     - packagebundlecontrollers
   sideEffects: None
+- admissionReviewVersions:
```
lewisdiamond, line 29: This is in PR #46

pkg/signature/manifest.go
```diff
@@ -18,13 +18,14 @@ import (
 )
 
 const (
-	PublicKey               = "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAETP8OUc6rTZHJs98X1aJfDIO0BXihHnSBDJhacdxZwk9RVzq28OIxQSVXXhD5ATEqWcNSWLnCG/GrZY9W2NfoMw=="
```
lewisdiamond, line 21: Updating the key to reflect the new key post `AddOns` -> `Packages` rename which also resulted in a new key name.

pkg/signature/manifest.go
```diff
@@ -133,20 +134,9 @@ func GetDigest(manifest Manifest, domain Domain) (digest [32]byte, yml []byte, e
 	return digest, yml, err
 }
 
+//See ./testdata/sign_file.sh for a shell script implementation.
+//This here differs in that it normalizes quoting while the shell script doesnt (yet).
 func ValidateSignature(manifest Manifest, domain Domain) (valid bool, digest [32]byte, yml []byte, err error) {
-	// Shell equivalent
```
lewisdiamond, line 137: We should not keep this comment up-to-date, the shell script is a better reference.

api/v1alpha1/packagebundle_webhook.go
```diff
@@ -50,10 +55,18 @@ func (r *PackageBundle) ValidateDelete() error {
 }
 
 func (r *PackageBundle) validate() error {
-	valid, digest, yml, err := signature.ValidateSignature(r, signature.EksaDomain)
+	//TODO Turn this into fields in a new CRD used for signature validation configuration
```
TerryHowe, line 58: We need a new CRD? I thought this was going in PackageBundleController

api/v1alpha1/packagebundle_webhook.go
```diff
@@ -50,10 +55,18 @@ func (r *PackageBundle) ValidateDelete() error {
...
+	keyOverride := os.Getenv(PublicKeyEnvVar)
+	var domain signature.Domain
+	if keyOverride != "" {
+		domain = signature.Domain{Name: signature.DomainName, Pubkey: keyOverride}
+	} else {
+		domain = signature.EksaDomain
+	}
+	valid, digest, yml, err := signature.ValidateSignature(r, domain)
```
ewollesen, line 66: Your logic here is a little muddied. I'd prefer a more straightforward approach (github won't let me make this as a suggestion, I guess because it touches lines that you didn't modify..?)

```go
valid, digest, yml, err := signature.ValidateSignature(r, domain)
if err != nil {
    return err
}
if !valid {
    return fmt.Errorf("invalid signature for configured public key %q", domain.Pubkey)
}
return nil
```
This deals with one error at a time, and makes it explicit that if you've reached the end of the function, you've succeeded. Alternatively, you _could_ configure `ValidateSignature` to return an error if it fails to validate. 

Additionally, it's generally a bad idea in go to both log and return an error. The top-level caller should log if that makes sense, but it's best to leave that decision to them. If you're going to ignore an error (cuz it's recoverable or whatever), then it can make sense to log that.

api/v1alpha1/webhook_suite_test.go
```diff
@@ -24,11 +24,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/eks-anywhere-packages/api/v1alpha1"
```
ewollesen, line 27: Looks like some things need linting?

api/v1alpha1/webhook_suite_test.go
```diff
@@ -134,6 +178,11 @@ var _ = BeforeSuite(func() {
 		conn.Close()
 		return nil
 	}).Should(Succeed())
+
+	ns := corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: "eksa-packages"},
```
ewollesen, line 183:
```suggestion
		ObjectMeta: metav1.ObjectMeta{Name: bundle.ActiveBundleNamespace},
```

ewollesen (changes requested):

api/v1alpha1/packagebundle_webhook.go
lewisdiamond: Ideally we'd use a new CRD and completely extract all the signature validation stuff into its own thing that can live without the eksa packages controller. That way people could use that for other types of validations. TBD

Now this may not get much adoption since we're also planning to provide that functionality via OPA.

api/testdata/bundle_two.yaml.signed
```diff
@@ -0,0 +1,30 @@
+apiVersion: packages.eks.amazonaws.com/v1alpha1
```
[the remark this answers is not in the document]
lewisdiamond: We could but I have not come across any incompatibilities so far. If we hit an issue due to that it'll be an easy decision.

api/v1alpha1/packagebundle_webhook.go
```diff
@@ -50,10 +55,18 @@ func (r *PackageBundle) ValidateDelete() error {
...
 func (r *PackageBundle) validate() error {
-	valid, digest, yml, err := signature.ValidateSignature(r, signature.EksaDomain)
+	//TODO Turn this into fields in a new CRD used for signature validation configuration
+	keyOverride := os.Getenv(PublicKeyEnvVar)
+	var domain signature.Domain
+	if keyOverride != "" {
+		domain = signature.Domain{Name: signature.DomainName, Pubkey: keyOverride}
+	} else {
+		domain = signature.EksaDomain
+	}
```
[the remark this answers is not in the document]
lewisdiamond: Yes what you suggested is also my preference. From this: https://go.dev/doc/faq#Does_Go_have_a_ternary_form I thought it was the "idiomatic" way, although it lacks context (the variable declaration is missing). If more than one person in the team prefers the suggested approach it's enough for me.

api/v1alpha1/packagebundle_webhook.go
lewisdiamond: I normally avoid multiple return points in functions since it tends not to age well. That doesn't seem like a Go idiom though. I've done it in other functions ( I cringe a bit when looking at them ), I could do it here too.

I prefer if `ValidateSignature` returns an error in the case of an actual error and a boolean for validity to avoid trying to match on the error type or message. It allows for the `apilog.Info` which is there to provide details which are not to be returned to the caller and displayed by the `kubectl` output. The log will be available via `kubectl logs` whereas the `error` is printed to `kubectl`'s stdout.

api/v1alpha1/webhook_suite_test.go
lewisdiamond: Yea it seems like I need to configure something for imports in my environment.

api/v1alpha1/packagebundle_webhook.go
ewollesen: > I normally avoid multiple return points in functions since it tends not to age well. That doesn't seem like a Go idiom though. I've done it in other functions ( I cringe a bit when looking at them ), I could do it here too.

You're correct, it's not a go idiom. It's very common in go to return, especially when there's an error. I found it a little off-putting at first, but have since come to appreciate it, as they are generally arranged so as to minimize the amount of info you have to keep in your head at any one time. I believe the egregious uses of return in the middle of a function are for non-error cases. It's a smell that your function is doing more than one thing.

> I prefer if `ValidateSignature` returns an error in the case of an actual error and a boolean for validity to avoid trying to match on the error type or message. It allows for the `apilog.Info` which is there to provide details which are not to be returned to the caller and displayed by the `kubectl` output. The log will be available via `kubectl logs` whereas the `error` is printed to `kubectl`'s stdout.

That all seems well-reasoned. I have some reservations about logging in that way, but it doesn't in any way affect program correctness so 👍🏻 and I'll try to keep my eye on it, and see how it works out.

api/v1alpha1/webhook_suite_test.go
ewollesen: Yeah, I've never used as many separate tools to do this, as we use here, but this is what I'm using presently:

```sh
GOPATH="${HOME}/go"
local="$(go list -m)"
gofmt -s -w "$@"
goimports -w -local "$local" "$@"
gci -w -local "$local" "$@"
```

api/v1alpha1/packagebundle_webhook.go
```diff
@@ -50,10 +55,19 @@ func (r *PackageBundle) ValidateDelete() error {
 }
 
 func (r *PackageBundle) validate() error {
-	valid, digest, yml, err := signature.ValidateSignature(r, signature.EksaDomain)
-	if err == nil && !valid {
+	//TODO Turn this into fields in a new CRD used for signature validation configuration
+	keyOverride := os.Getenv(PublicKeyEnvVar)
```
ewollesen, line 59: This might be overkill, but in the name of visibility and reproducibility... What if we inserted this value at build time?

Makefile (pseudo):
```
build:
    go build -ldflags "-X api.v1alpha1.keyOverride=$(data here, from env or catting a file)" .
```
The advantage of this version, is that the binary contains the public key, it could be identified with `strings` or the like, and there's no doubt about which key was used. With the env var approach, the key could change with each run of the code. We could have the dev/testing key in the code itself, and override with prod/staging at build time.

I guess another alternative is to, at build time, generate a file that defines the public key as a constant. That's in some ways even better, since as a constant, it can't be changed. Though a non-exported var also can't be changed (by code outside of the package that we control at least)...

Anyway, I'm not asking for a change here, more just brainstorming possible improvements.

api/v1alpha1/webhook_suite_test.go
lewisdiamond: Yea I actually had to downgrade gci too I believe.

api/v1alpha1/packagebundle_webhook.go
lewisdiamond: Ultimately we want to allow users to define trusted keys and domains so all of this would make its way to a CRD. If we keep a hard-coded copy of the key, we could use what you propose to override it in different build stages. The downside is however that we'd end up with non-matching digests across environments. We would not be able to simply promote a build.

ewollesen (approved):
/lgtm

PR #49, opened 2022-03-22
Add a default packagebundlecontroller to the chart

lewisdiamond:
Also move CRDs to the `crds/` folder to avoid being unable to deploy a
packagebundlecontroller within the chart.


TerryHowe (approved):
/lgtm

I tried to do this and i thought the file name being crd would be enough to trigger helm to load it first. Didn't work. Cool you got this working. Need to add a default packageController CR some day

PR #54, opened 2022-03-23
Install by version and upgrade all on new releases

lewisdiamond:

### DIFF
--- a/api/v1alpha1/packagebundle.go
+++ b/api/v1alpha1/packagebundle.go
@@ -6,6 +6,7 @@ import (
 )
 
 type PackageOCISource struct {
+	Version    string `json:"version"`
 	Registry   string `json:"registry"`
 	Repository string `json:"repository"`
 	Digest     string `json:"digest"`
@@ -13,6 +14,7 @@ type PackageOCISource struct {
 
 const (
 	PackageBundleKind = "PackageBundle"
+	Latest            = "latest"
 )
 
 func (config *PackageBundle) MetaKind() string {
@@ -23,20 +25,22 @@ func (config *PackageBundle) ExpectedKind() string {
 	return PackageBundleKind
 }
 
-func (config *PackageBundle) FindSource(pkgName, pkgVersion string) (retSource PackageOCISource, version string, err error) {
+func (config *PackageBundle) FindSource(pkgName, pkgVersion string) (retSource PackageOCISource, err error) {
 	for _, pkg := range config.Spec.Packages {
 		if pkg.Name == pkgName {
 			source := pkg.Source
 			for _, version := range source.Versions {
-				if version.Name == pkgVersion || version.Digest == pkgVersion {
-					retSource = PackageOCISource{Registry: source.Registry, Repository: source.Repository, Digest: version.Digest}
-					return retSource, version.Name, nil
+				//We do not sort before getting `latest` because there will be only a single version per release in normal cases. For edge cases which may require multiple
+				//versions, the order in the file will be ordered according to what we want `latest` to point to
+				if version.Name == pkgVersion || version.Digest == pkgVersion || pkgVersion == Latest {
+					retSource = PackageOCISource{Registry: source.Registry, Repository: source.Repository, Digest: version.Digest, Version: version.Name}
+					return retSource, nil
 				}
 			}
 		}
 	}
 
-	return retSource, "", fmt.Errorf("package not found: %s @ %s", pkgName, pkgVersion)
+	return retSource, fmt.Errorf("package not found in bundle (%s): %s @ %s", config.ObjectMeta.Name, pkgName, pkgVersion)
 }
 
 func (s PackageOCISource) AsRepoURI() string {
--- a/api/v1alpha1/packagebundle_test.go
+++ b/api/v1alpha1/packagebundle_test.go
@@ -11,45 +11,98 @@ import (
 
 func TestPackageBundle_Find(t *testing.T) {
 	var err error
-	sut := api.PackageBundle{
-		Spec: api.PackageBundleSpec{
-			Packages: []api.BundlePackage{
-				{
-					Name: "eks-anywhere-test",
-					Source: api.BundlePackageSource{
-						Registry:   "public.ecr.aws/l0g8r8j6",
-						Repository: "eks-anywhere-test",
-						Versions: []api.SourceVersion{
-							{
-								Name:   "v0.1.0",
-								Digest: "sha256:eaa07ae1c06ffb563fe3c16cdb317f7ac31c8f829d5f1f32442f0e5ab982c3e7",
-							},
+	givenBundle := func(versions []api.SourceVersion) api.PackageBundle {
+		return api.PackageBundle{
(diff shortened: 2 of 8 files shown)


api/v1alpha1/packagebundle.go
```diff
@@ -23,20 +25,22 @@ func (config *PackageBundle) ExpectedKind() string {
...
 	for _, pkg := range config.Spec.Packages {
 		if pkg.Name == pkgName {
 			source := pkg.Source
 			for _, version := range source.Versions {
-				if version.Name == pkgVersion || version.Digest == pkgVersion {
-					retSource = PackageOCISource{Registry: source.Registry, Repository: source.Repository, Digest: version.Digest}
-					return retSource, version.Name, nil
+				//We do not sort before getting `latest` because there will be only a single version per release in normal cases. For edge cases which may require multiple
+				//versions, the order in the file will be ordered according to what we want `latest` to point to
+				if version.Name == pkgVersion || version.Digest == pkgVersion || pkgVersion == Latest {
```
ewollesen, line 35: Not asking you to change this, but should we be doing a case-insensitive comparison? At least for the digest, case definitely doesn't matter, and I think we can make a case that case shouldn't matter in either of the other cases as well.

Are there any cases where case would matter for this purpose? I can't think of any. Also, at least in the case of a digest, I can imagine someone specifying the digest in whatever case we're not expecting.

api/v1alpha1/packagebundle_test.go
```diff
@@ -11,45 +11,96 @@ import (
...
 
-	actual, version, err = sut.FindSource("eks-anywhere-test", "sha256:eaa07ae1c06ffb563fe3c16cdb317f7ac31c8f829d5f1f32442f0e5ab982c3e7")
+	actual, err = sut.FindSource("eks-anywhere-test", "sha256:eaa07ae1c06ffb563fe3c16cdb317f7ac31c8f829d5f1f32442f0e5ab982c3e7")
 	assert.NoError(t, err)
 	assert.Equal(t, expected, actual)
-	assert.Equal(t, expectedVersion, version)
 
-	expectedErr := "package not found: Bogus @ bar"
-	_, _, err = sut.FindSource("Bogus", "bar")
+	expectedErr := "package not found in current active bundle: Bogus @ bar"
```
ewollesen, line 53:
```suggestion
	expectedErr := "package not found in active bundle (): Bogus @ bar"
```

controllers/package_controller.go
```diff
@@ -124,8 +144,19 @@ func (r *PackageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		targetVersion := managerContext.Package.Spec.PackageVersion
+		if targetVersion == "" {
+			targetVersion = api.Latest
+		}
+		pkgName := managerContext.Package.Spec.PackageName
+		managerContext.Source, err = bundle.FindSource(pkgName, targetVersion)
+		managerContext.Package.Status.TargetVersion = printableTargetVersion(managerContext.Source, targetVersion)
 		if err != nil {
+			//TODO add a link to documentation on how to make a bundle active.
+			managerContext.Package.Status.Detail = fmt.Sprintf("Package %s@%s is not in the current active bundle. Did you forget to activate the new bundle?", pkgName, targetVersion)
```
ewollesen, line 156:
```suggestion
			managerContext.Package.Status.Detail = fmt.Sprintf("Package %s@%s is not in the active bundle (%s). Did you forget to activate the new bundle?", bundle.Version, pkgName, targetVersion)
```

I wouldn't output the second sentence unless I knew there was a newer bundle on the system. I feel like error messages that are misleading are worse than terse ones.

controllers/package_controller_test.go
```diff
@@ -165,6 +170,89 @@ func TestReconcile(t *testing.T) {
...
+			DoAndReturn(fn)
+
+		tf.bundleManager.FakeActiveBundle = tf.mockBundle()
+
+		testErr := errors.New("status update test error")
+		status := tf.mockStatusWriter()
+
+		pkg.Spec.PackageVersion = "2.0.0"
+		pkg.Status.TargetVersion = "2.0.0"
+		pkg.Status.Detail = fmt.Sprintf("Package %s@%s is not in the current active bundle. Did you forget to activate the new bundle?", pkg.Spec.PackageName, pkg.Spec.PackageVersion)
```
ewollesen, line 189:
```suggestion
		pkg.Status.Detail = fmt.Sprintf("Package %s@%s is not in the active bundle (). Did you forget to activate the new bundle?", pkg.Spec.PackageName, pkg.Spec.PackageVersion)
```
Heh, there's no bundle version in `tf.mockBundle()`

api/v1alpha1/packagebundle.go
```diff
@@ -23,20 +25,22 @@ func (config *PackageBundle) ExpectedKind() string {
...
+					retSource = PackageOCISource{Registry: source.Registry, Repository: source.Repository, Digest: version.Digest, Version: version.Name}
+					return retSource, nil
 				}
 			}
 		}
 	}
 
-	return retSource, "", fmt.Errorf("package not found: %s @ %s", pkgName, pkgVersion)
+	return retSource, fmt.Errorf("package not found in current active bundle: %s @ %s", pkgName, pkgVersion)
```
ewollesen, line 43:
```suggestion
	return retSource, fmt.Errorf("package not found in active bundle (%s): %s @ %s", config.Spec.Version, pkgName, pkgVersion)
```

I've made a couple of suggestions in line with this one, but just via the web interface, so I may well have goofed something. Hopefully they're more useful than not though.

api/v1alpha1/packagebundle_test.go
```diff
@@ -11,45 +11,96 @@ import (
...
+	givenBundle := func(versions []api.SourceVersion) api.PackageBundle {
+		return api.PackageBundle{
+			Spec: api.PackageBundleSpec{
+				Packages: []api.BundlePackage{
+					{
+						Name: "eks-anywhere-test",
+						Source: api.BundlePackageSource{
+							Registry:   "public.ecr.aws/l0g8r8j6",
+							Repository: "eks-anywhere-test",
+							Versions:   versions},
```
ewollesen, line 23:
```suggestion
							Versions:   versions,
					        },
```
Very much a nit, but....

ewollesen (approved):
Just small ideas/tweaks from me. I can be persuaded to merge as-is though.

api/v1alpha1/packagebundle_test.go
lewisdiamond: Isn't that supposed to get linted? I did run it with `--fix`

ewollesen: I don't think so, strangely enough... There are occasions where you might leave it this way, e.g.:
```go
foo := myType{ foo: "bar",
    ack: "gar"}
```
Notice the assignments start on the first line. I don't love this style, but sometimes it makes sense.

lewisdiamond:
> @lewisdiamond: you cannot LGTM your own PR.
Well seems like you're alive.

PR #55, opened 2022-03-24
Add helm chart configuration to bundle

TerryHowe:
Add configurations to bundle versions so for each helm chart, we can have the associated valid configurations


api/v1alpha1/packagebundle_types.go
```diff
@@ -83,6 +83,9 @@ type SourceVersion struct {
 
 	// Images is a list of images used by this version of the package
 	Images []VersionImages `json:"images,omitempty"`
+
+	// Configurations is a list of configurations used by this version of the package
```
ewollesen, line 87: I wouldn't mind an added bit of commentary for how these are used.

ewollesen (approved):

api/v1alpha1/packagebundle_types.go
TerryHowe: Done

ewollesen (approved):
/lgtm

TerryHowe:
> This is for overriding values in the helm charts? Or is this for something else.

These are the config values for the our package custom resources. Should map to a subset of the `values.yaml`

I wonder why this hasn't merged.

PR #56, opened 2022-03-24
Helm bugbash changes  + Public key env var

jonahjon:

*Issue #, if available:*

*Description of changes:*

Using Dev Gallery Address for Bug Bash, and local testing since the dev image is now live.  https://gallery.ecr.aws/l0g8r8j6/eks-anywhere-packages

Also added in the Public Key from Dev account, and the ENV Var to values.

### DIFF
--- a/charts/eks-anywhere-packages/values.yaml
+++ b/charts/eks-anywhere-packages/values.yaml
@@ -67,12 +67,15 @@ controller:
   repository: "eks-anywhere-packages"
   # -- Controller image tag, or sha sum.
   tag: "v0.1.2-eks-a-v0.0.0-dev-build.1944"
+  # -- Whether to turn on Webhooks for the controller image
+  enableWebhooks: "true"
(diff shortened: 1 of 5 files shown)


charts/eks-anywhere-packages/values.yaml
```diff
@@ -66,11 +66,13 @@ controller:
   # -- Controller repository name.
   repository: "eks-anywhere-packages"
   # -- Controller image tag, or sha sum.
-  tag: "latest"
+  tag: "v0.1.2-eks-a-v0.0.0-dev-build.1944"
   # -- Additional environment variables for the controller pod.
   env:
     - name: ENABLE_WEBHOOKS
       value: "true"
+    - name: EKSA_PUBLIC_KEY
```
lewisdiamond, line 74: This isn't needed as the key is hard coded and the environment variable is only needed to override it.

lewisdiamond (approved):

PR #58, opened 2022-03-25
Change the package.Spec.Config field to a string

lewisdiamond:
The string will be parsed as YAML and used as values for packages.


api/testdata/test.yaml
```diff
@@ -5,7 +5,11 @@ metadata:
...
-    secret.bar: eolus
-    pvc.storage: 1Mi
+  config: |
+    config:
+      - foo: windom
+      - baz: baz
+    secret:
+      bar: eolus
+    pvc:
+      storage: 1Mi
```
TerryHowe, line 15: I haven't looked at the code in general, but I like the general idea of this although I think we should attempt to keep the yaml flat. The helm best practices  of keeping values.yaml flat as possible I think also would help our situation

lewisdiamond: We'd have to try to enforce that on each chart we produce. In general most charts keep it only 1-3 levels deep.

TerryHowe (approved):

api/v1alpha1/package.go
```diff
@@ -19,32 +17,6 @@ func (config *Package) ExpectedKind() string {
...
-				}
-			}
-			newMap := make(map[string]interface{})
-			subMap[name] = newMap
-			subMap = newMap
-		}
-		subMap[key] = v
-	}
-	return mapInterfaces, nil
+	err = yaml.Unmarshal([]byte(config.Spec.Config), &mapInterfaces)
```
ewollesen, line 20: We should probably wrap this error to add a little context.

PR #72, opened 2022-03-30
Use the dev public key for now

lewisdiamond:
For the bug bash, we use the test/dev public key.

### DIFF
--- a/charts/eks-anywhere-packages/values.yaml
+++ b/charts/eks-anywhere-packages/values.yaml
@@ -72,7 +72,9 @@ controller:
   # -- Additional environment variables for the controller pod.
   # - name: EKSA_PUBLIC_KEY
   #   value: ""
-  env: []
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

charts/eks-anywhere-packages/values.yaml
```diff
@@ -72,7 +72,9 @@ controller:
   # -- Additional environment variables for the controller pod.
   # - name: EKSA_PUBLIC_KEY
   #   value: ""
-  env: []
```
TerryHowe, line 75: Any particular reason this is an env here rather than a regular value that is an env on the deployment?

TerryHowe: This does not hide the implementation and will not work if we change the implementation

PR #86, opened 2022-04-06
New command: sign-object

ewollesen:
go run ./cmd/sign-object --help

It has a different take on how to generate the signature from what we
currently have.

The current solution uses github.com/itchyny/gojq (a Go-library that
implements jq). I've found that operations like we're performing are
very difficult with jq. The queries are brittle and hard to read (even
harder to modify).

I started looking to see if we could use Kustomize to patch in a
signature. However, Kustomize isn't flexible enough (this is a long
story).

So for this new bit of functionality, I've gone to the library that
underlies Kustomize, jsonpatch (github.com/evanphx/json-patch). I feel
like this is more resilient, though it's definitely more lines of
code. I think this will be more maintainable and composable in the long
run. Those familiar with Kustomize should have an easier time wrapping
their heads around how this code works. Also, if, in the future, the
limitations that prevent us from using Kustomize directly are removed,
we might be able to simply do that, which would simplify a chunk of
this.

I've not functionally modified the existing signature package, though I
did perform a few tweaks that were either sanitary (extracting
constants, etc) or just made the existing code a little easier to
read/use (moving from [32]byte to []byte).

If this approach is deemed worthwhile, I would like to expand to using
its approach in the signature package. I'd also be writing a bunch of
tests of course.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cmd/sign-object/main.go
+++ b/cmd/sign-object/main.go
@@ -1,0 +1,124 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"sigs.k8s.io/yaml"
+
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+	"github.com/aws/eks-anywhere-packages/pkg/signable"
+)
+
+const (
+	defaultKey    = "private.ec.key"
+	defaultOutput = "-"
+)
+
+func main() {
+	var keyFile string
+	var outputFile string
+
+	rootCmd := &cobra.Command{
+		Use:   "sign-object -s filename [-o object.signed.yaml] [object.yaml]",
+		Short: "Sign a kubernetes object file",
+		Long: `Sign a kubernetes object YAML file.
+
+The signing key is expected to be PEM-encoded.
+
+If no bundle file is supplied, reads from stdin.
+
+If no output file is specified, writes to stdout.
+`,
+		Args: cobra.MaximumNArgs(1), // a kubernetes object file for input
+		RunE: sign,
+	}
+	rootCmd.Flags().StringVarP(&keyFile, "signing-key", "s", defaultKey,
+		"private key filename for signing the package bundle")
+	rootCmd.Flags().StringVarP(&outputFile, "output", "o", defaultOutput,
+		"output bundle filename")
+	if err := rootCmd.MarkFlagRequired("signing-key"); err != nil {
+		log.Fatal(err)
+	}
+	if err := rootCmd.Execute(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+var DefaultGroup = api.GroupVersion.Group
+var DefaultVersion = api.GroupVersion.Version
+
+func sign(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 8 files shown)


ewollesen:
Here's an example of it in action:
```sh
$ sign-object -s pkg/signature/testdata/private.ec.key api/testdata/bundle_one.yaml
apiVersion: packages.eks.amazonaws.com/v1alpha1
kind: PackageBundle
metadata:
  annotations:
    eksa.aws.com/signature: MEUCIDBLkkCKmlXfEe/Eidzfz8d+dWIlwrqrInJFcioXxwixAiEAsIhv9eT+2TynhKcOVGa3xhXaNJKbFI3q71Vr6pK0uIw=
  name: v1-21-1001
  namespace: eksa-packages
spec:
  kubeVersion: v1.21
  packages:
  - name: Test
    source:
      registry: public.ecr.aws/l0g8r8j6
      repository: eks-anywhere-test
      versions:
      - digest: sha256:64ea03b119d2421f9206252ff4af4bf7cdc2823c343420763e0e6fc20bf03b68
        name: v0.1.1-8b3810e1514b7432e032794842425accc837757a-helm
  - name: Flux
    source:
      registry: public.ecr.aws/j0a1m4z9
      repository: fluxcd/flux2
      versions:
      - digest: sha256:5fb61d0b650483851ce3892e30f1d2c4c1372aafd51be24f37bbb0056db04012
        name: v0.23.0-0b585d5986a1cfcee67b7a956eeef687f00ae468-helm
  - name: Harbor
    source:
      registry: public.ecr.aws/y8n6a2y0
      repository: harbor/harbor-helm
      versions:
      - digest: sha256:423eb28b0586376f4d1d30b492370a4e215f2b0210a587fff6f358efc41dda4c
        name: v2.4.1-f47374093e8a9c48aca8c7a7f06ae185eb7506f3-helm
```

You can easily sign any k8s object that we've defined in `api/v1alpha1`:

```sh
$ sign-object -s pkg/signature/testdata/private.ec.key api/testdata/harbor.yaml 
apiVersion: packages.eks.amazonaws.com/v1alpha1
kind: Package
metadata:
  annotations:
    eksa.aws.com/signature: MEUCIQDKQdyFX7iamRzEuZ/ZmRFbEfbCh6cvfBE+x7BRPuGv5QIgOtUZt3NYHDBXjShg7sj8HHF4ayQzYOd/cpAbLwHKNco=
  name: my-harbor
  namespace: eksa-packages
spec:
  packageName: Harbor
  packageVersion: v2.4.1-f47374093e8a9c48aca8c7a7f06ae185eb7506f3-helm
```
```sh
$ sign-object -s pkg/signature/testdata/private.ec.key api/testdata/test.yaml 
apiVersion: packages.eks.amazonaws.com/v1alpha1
kind: Package
metadata:
  annotations:
    eksa.aws.com/signature: MEUCIBZyfkEyWn+EdnLk55DppT+VwwNnBrtKg3H134JQ+2GyAiEAvbJcQ8xL4dHCvU4I2As5sg66JSDbKPUfIW97uVj5cU0=
  name: my-test
  namespace: eksa-packages
spec:
  config: |
    config:
      - foo: windom
      - baz: baz
    secret:
      bar: eolus
    pvc:
      storage: 1Mi
  packageName: Test
  packageVersion: v0.1.0
```

TerryHowe (approved):

jonahjon:
I like this, I'll change the signing job to use this once it merges.

jonahjon:
Would it be possible to add a Makefile with build commands for this so it can be used in the pipelines?

ewollesen:
stale

PR #98, opened 2022-04-12
Increase resources limit on controller

lewisdiamond:
We've had reports of the controller getting `OOMKilled`, increasing the limit will prevent that.

I've profiled the controller and I see total usage of the container go up to 101MiB, the controller itself doesn't report much heap usage however. More profiling will be done in the future.


TerryHowe (approved):
/lgtm

Issue #83, opened 2022-04-04
Package configuration modifications are not immediately applied


lewisdiamond:
1. Install a package
2. Modify the package `config` field
3. Apply the modified package

Expected: The package is updated to reflect the config
Actual: Nothing happens, waits for an actual version upgrade.

PR #102, opened 2022-04-19
Apply config changes to existing packages

ewollesen:
Closes: https://github.com/aws/eks-anywhere-packages/issues/83

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/packages/manager.go
+++ b/pkg/packages/manager.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"sigs.k8s.io/yaml"
 
 	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
 	"github.com/aws/eks-anywhere-packages/pkg/driver"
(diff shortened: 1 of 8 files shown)


pkg/packages/manager.go
```diff
@@ -77,6 +78,32 @@ func processInstalled(mc *ManagerContext) bool {
...
+			mc.RequeueAfter = retryShort
+			return true
+		}
+
+		needs, err := mc.PackageDriver.IsConfigChanged(mc.Ctx, mc.Package.Name,
+			newValues)
+		if err != nil {
+			mc.Log.Error(err, "checking necessity of reconfiguration")
+			mc.Package.Status.Detail = err.Error()
+			mc.RequeueAfter = retryShort
```
TerryHowe, line 98: This could probably be long or very long, we should be notified if they update/fix

pkg/packages/manager.go
```diff
@@ -77,6 +78,32 @@ func processInstalled(mc *ManagerContext) bool {
...
+			return true
+		}
+		if needs {
+			mc.Log.Info("configuration change detected, upgrading")
+			mc.Package.Status.State = api.StateUpdating
+			mc.RequeueAfter = retryShort
+			return true
```
TerryHowe, line 105: tests maybe?

pkg/packages/manager.go
ewollesen: I re-read what I could find about this in kubebuilder, and am now thinking that `retryNever` should be used a lot more frequently than we are. Let me know if you disagree or understand something different.

TerryHowe: Event based is better than polling, so `retryNever` makes sense. I guess in the case of an error though, we might want to retry *sometime* in case the problem was ephemeral

ewollesen: My understanding is that k8s will queue the resources periodically, even if not requested. Let me see if I can find the docs on that.

ewollesen: I can't find a solid reference to that right now. Will set to verylong.

pkg/packages/manager.go
ewollesen: I've added some in the latest push.

TerryHowe: Be nice to have a test a least for the "needs" case

TerryHowe (approved):
/lgtm

PR #115, opened 2022-04-22
Add .igntore director to ignore like eks-anywhere

TerryHowe:


jonahjon (approved):

ewollesen:
I guess this is fine, but `.git/config/exclude` already exists for this purpose..? So do we need this?

PR #122, opened 2022-04-25
Remove print from bundle manager

TerryHowe:
Kind of painful to remove this, but this print is deceptive because no relationship has been established between the bundles at this point. We can pull this out of git history if we need, but it is confusing at this point.


ewollesen (approved):
/lgtm

PR #129, opened 2022-04-27
Add bundle client

TerryHowe:
Add basic bundle client


pkg/bundle/bundle_client.go
```diff
@@ -0,0 +1,63 @@
+package bundle
```
ewollesen, line 1: I'd suggest renaming the file to `client.go`, so as to avoid stuttering.
> A package’s name provides context for its contents. For instance, the [bytes package](https://go.dev/pkg/bytes/) from the standard library exports the Buffer type. On its own, the name Buffer isn’t very descriptive, but when combined with its package name its meaning becomes clear: bytes.Buffer. If the package had a less descriptive name, like util, the buffer would likely acquire the longer and clumsier name util.BytesBuffer.

-- https://go.dev/blog/organizing-go-code

controllers/package_controller.go
```diff
@@ -80,12 +82,14 @@ func RegisterPackageReconciler(mgr ctrl.Manager) (err error) {
 	}
 	puller := artifacts.NewRegistryPuller()
 	bundleManager := bundle.NewBundleManager(log, discovery, puller)
```
ewollesen, line 84: I wonder if the BundleManager should be passed a BundleClient instead of a k8s client for its `IsActive` and `ActiveBundle` methods. Or alternatively at instantiation (though I think that might break the abstraction we were going for.)

In any event, the larger point here, that I'd like to get your feedback on is this: We now have two different constructs that are able to retrieve the active bundle. I'd prefer there to be one place to get that information. Meaning that the BundleManager should probably have its IsActive and ActiveBundle methods removed... or something?

pkg/bundle/bundle_client.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+)
+
+type BundleClient interface {
```
ewollesen, line 13: As above, I don't see why this couldn't just be `Client`, Mostly it would be referenced as `bundle.Client` anyway.

ewollesen (commented):
I mainly just have questions. Not opposed to merging.

controllers/package_controller.go
TerryHowe: Agreed, but as a follow-up. I needed to make this PR as small as possible.

pkg/bundle/bundle_client.go
TerryHowe: Yes, good idea.

ewollesen:
Glad to review again if that works for you, or you can merge this and handle in a follow up.

ewollesen (commented):
/lgtm

ewollesen (approved):

PR #134, opened 2022-04-27
Update install workflow to support Go v1.18

ivyostosh:
Command `go get` can no longer be used to build packages starting `Go v1.18` (see [deprecation notice](https://go.dev/doc/go-get-install-deprecation)). Kustomize version also needs to be updated since `go install` doesn't support exclude directives (see [issue](https://github.com/kubernetes-sigs/kustomize/issues/3618)). The change has been tested on both `Go v1.17` and `Go v1.18` to ensure the Makefile works properly.

*Issue #, if available:* N/A

*Description of changes:* Update `go get` to `go install`, and Kustomize version to `v4.5`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ kubetest2: ## Download kubetest2 locally if necessary.
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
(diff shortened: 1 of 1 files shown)


TerryHowe:
/hold
We all need to run 17.9 I'm afraid

ewollesen:
> /hold We all need to run 17.9 I'm afraid

We spoke some at standup, and I've verified with @ivyostosh that all uses of `go-get-tool` succeeds (which was where there were problems in the past). Since this works just fine for 1.17.9 and will work in the future with 1.18, I think this is good to merge.

/unhold
/lgtm

ewollesen (approved):

PR #135, opened 2022-04-28
Update package bundle image

ivyostosh:
Update the name of `hello-eks-anywhere` repository to match an existing version on `public.ecr.aws/l0g8r8j6/hello-eks-anywhere`.

*Issue #, if available:* `v0.1.1-8b3810e1514b7432e032794842425accc837757a-helm` is not a valid name for `hello-eks-anywhere`.

*Description of changes:* Update the name to an existing version on `public.ecr.aws/l0g8r8j6/hello-eks-anywhere`.

### DIFF
--- a/api/testdata/bundle_one.yaml
+++ b/api/testdata/bundle_one.yaml
@@ -11,7 +11,7 @@ spec:
         registry: public.ecr.aws/l0g8r8j6
         repository: hello-eks-anywhere
         versions:
-          - name: v0.1.1-8b3810e1514b7432e032794842425accc837757a-helm
+          - name: 0.1.1_b41df7249b38548e535531d2e971ba85bb374434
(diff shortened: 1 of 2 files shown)


TerryHowe:
Are you able to reproduce that presubmit error with `make test` ?

jonahjon:
> Are you able to reproduce that presubmit error with `make test` ?

The error looks unrelated to the change, since she's just changing the tag. Could be wrong though

ewollesen:
> > Are you able to reproduce that presubmit error with `make test` ?
> 
> The error looks unrelated to the change, since she's just changing the tag. Could be wrong though

I only looked briefly, but I'd disagree. The errors are with the signing. Since this bundle changed, the signature that goes with it also needs to change. I believe running `make test` _should_ regenerate those signed artifacts, which should then be added to this commit.

ivyostosh:
It looks like we can successfully pass all checks after an update to the `.signed` file.

ewollesen (approved):

Issue #136, opened 2022-04-28
Need method to download any bundle by tag


TerryHowe:
If someone deletes the active bundle, we should download it again.

- [x] We need this method https://github.com/aws/eks-anywhere-packages/blob/main/pkg/bundle/manager.go#L145 the LatestBundle method, but the download part needs to be pulled out and a public DownloadBundle method created.
- [x] probably need a method to get the name of the active bundle
- [x] If the bundle controller detects the active bundle was deleted https://github.com/aws/eks-anywhere-packages/blob/main/controllers/packagebundle_controller.go#L106 currently it is ignored, but it should attempt to download the bundle again
- [x] Use the k8s client interface to recreate the bundle

PR #141, opened 2022-04-29
Create DownloadBundle method

ivyostosh:
Extract `DownloadBundle` out of the `LatestBundle` method so that the `DownloadBundle` method can be re-used for bundle with different tags. Separate unit test cases have been added and all passed.

*Issue #, if available:* https://github.com/aws/eks-anywhere-packages/issues/136

*Description of changes:* Extract `DownloadBundle` out of the `LatestBundle` method, and add testing cases.


pkg/bundle/manager.go
```diff
@@ -152,13 +156,24 @@ func (m *bundleManager) LatestBundle(ctx context.Context, baseRef string) (
 	tag := "latest"
 	ref := fmt.Sprintf("%s:%s-%s", baseRef, kubeVersion, tag)
 
+	bundle, err := m.DownloadBundle(ctx, ref)
+
+	if err != nil {
+		return nil, fmt.Errorf("downloading package bundle: %s", err)
+	}
+
+	return bundle, nil
```
TerryHowe, line 165:
```suggestion
	return m.DownloadBundle(ctx, ref)
```

TerryHowe (approved):
/lgtm

PR #148, opened 2022-05-03
Add a re-download functionality to Controller

ivyostosh:
Update the `Reconcile` function to support the scenario that if the bundle controller detects that the active bundle is deleted, the bundle controller will validate the active bundle by namespace and name, and re-download the bundle. The changes were tested locally and passed all testing.

*Issue #, if available:* https://github.com/aws/eks-anywhere-packages/issues/136

*Description of changes:* `Reconcile` function was updated to support the re-download logic. In order to support this logic, I also added a `GetActiveBundleNamespacedName` function in the `manager`, and updated `manager` `fake` functions. 

### DIFF
--- a/controllers/packagebundle_controller.go
+++ b/controllers/packagebundle_controller.go
@@ -110,7 +110,6 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		// the bundle controller will validate the active bundle by namespace
 		// and name, and re-download the bundle.
 		nn, err := r.bundleManager.GetActiveBundleNamespacedName(ctx, r.Client)
-
 		if err != nil {
 			r.Log.Info("Unable to get active bundle namespace and name",
 				"NamespaceName", nn)
--- a/controllers/packagebundle_controller.go
+++ b/controllers/packagebundle_controller.go
@@ -122,7 +122,7 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 			return ctrl.Result{}, nil
 		}
 
-		_, err = r.bundleManager.DownloadBundle(ctx, req.Name))
+		_, err = r.bundleManager.DownloadBundle(ctx, req.Name)
 
 		if err != nil {
 			r.Log.Error(err, "Active bundle deleted and failed to download", "bundle",
--- a/controllers/packagebundle_controller.go
+++ b/controllers/packagebundle_controller.go
@@ -105,7 +105,32 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		if client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{}, err
 		}
-		r.Log.Info("Bundle deleted (ignoring)", "bundle", req.NamespacedName)
+
+		// If the bundle controller detects that the active bundle is deleted,
+		// the bundle controller will validate the active bundle by namespace
+		// and name, and re-download the bundle.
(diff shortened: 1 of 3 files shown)


controllers/packagebundle_controller.go
```diff
@@ -105,7 +105,32 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+		if err != nil {
+			r.Log.Info("Unable to get active bundle namespace and name",
+				"NamespaceName", nn)
+			return ctrl.Result{}, nil
+		}
+
+		if nn.Namespace == req.Namespace && nn.Name == req.Name {
+			r.Log.Info("Bundle deleted", "bundle", req.NamespacedName)
+		}
```
TerryHowe, line 122:
```suggestion
		if nn.Namespace != req.Namespace || nn.Name != req.Name {
			r.Log.Info("Bundle deleted", "bundle", req.NamespacedName)
			return ctrl.Result{}, nil
		}
```

controllers/packagebundle_controller.go
```diff
@@ -105,7 +105,32 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+		ref := fmt.Sprintf("%s:%s", req.Namespace, req.Name)
+		_, err = r.bundleManager.DownloadBundle(ctx, ref)
```
TerryHowe, line 125: The parameter here I think is the tag in the registry which only is the name, not the namespace

```suggestion
		_, err = r.bundleManager.DownloadBundle(ctx, req.Name))
```

controllers/packagebundle_controller.go
```diff
@@ -105,7 +105,32 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+		if err != nil {
+			r.Log.Info("Unable to download deleted bundle", "bundle",
+				req.NamespacedName)
```
TerryHowe, line 129:
```suggestion
			r.Log.Error(err, "Active bundle deleted and failed to download", "bundle",
				req.NamespacedName)
```

controllers/packagebundle_controller.go
TerryHowe: Okay, actually I think there is a problem in general with DownloadBundle, but it should work as I suggest

ivyostosh: Let me follow up with you separately on the `DownloadBundle` method

controllers/packagebundle_controller.go
```diff
@@ -105,7 +105,32 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 			return ctrl.Result{}, err
 		}
-		r.Log.Info("Bundle deleted (ignoring)", "bundle", req.NamespacedName)
+
+		// If the bundle controller detects that the active bundle is deleted,
+		// the bundle controller will validate the active bundle by namespace
+		// and name, and re-download the bundle.
+		nn, err := r.bundleManager.GetActiveBundleNamespacedName(ctx, r.Client)
+
+		if err != nil {
```
ewollesen, line 114:
```suggestion
		nn, err := r.bundleManager.GetActiveBundleNamespacedName(ctx, r.Client)
		if err != nil {
```
Nit: it's customary to keep the error checking immediately following a call. It helps us recognize upon scanning the pattern of "make a call, check the error".

ivyostosh: Noted. Thanks.

TerryHowe (approved):
/lgtm

PR #150, opened 2022-05-03
Remove ActiveBundle method

TerryHowe:
This method was replaced by GetActiveBundle in the client


ivyostosh (approved):
Verified `ActiveBundle` is no longer needed and referenced. No format issue identified.

PR #151, opened 2022-05-03
Rename abc vars to pbc

TerryHowe:
These old abc var names are driving me nuts


ivyostosh (approved):
Not a blocker for this PR but [mango.go](https://github.com/aws/eks-anywhere-packages/blob/main/pkg/bundle/manager.go) also needs to be changed for the same name convention.

TerryHowe:
yeh on manager.go, but baby steps

PR #153, opened 2022-05-04
Restrict bundles to the eksa-packages namespace

TerryHowe:
The other PRs need to merge first


ivyostosh (approved):
/lgtm

I did a quick search to see if there are other use cases of `Client.List` as well in the repo as a consistency check, and confirmed these are the only two places that called `Client.List`.

ewollesen (approved):
/lgtm

PR #159, opened 2022-05-04
Move IsActive to client

TerryHowe:
Move the IsActive method to the client. This should be the last move.


ewollesen (approved):
Oof, I'm sad that I hadn't written any tests for you to move over. :(

/lgtm

PR #161, opened 2022-05-04
Fix active bundle handling

TerryHowe:
If a bundle is not active, we should make sure it is marked inactive.


pkg/bundle/manager_test.go
```diff
@@ -223,7 +223,7 @@ func TestUpdate(t *testing.T) {
 		}
 		bm := NewBundleManager(logr.Discard(), discovery, puller)
 
-		if assert.True(t, bm.Update(bundle, false, noBundles)) {
```
TerryHowe, line 226: "leaves state as-is (inactive)" should return false if the bundle was already inactive.

jonahjon (approved):

Issue #163, opened 2022-05-05
Prepare README for public preview


TerryHowe:
The README needs to be prepared for public preview

- [x] move the current contents to somewhere like `docs/developer.md`
- [x] follow this example https://github.com/aws/eks-anywhere
- [x] add a high level description of the project
- [x] badges, we need some stinking badges
- [x] development pointing to the former contents and high level description
- [x] security
- [x] license

ivyostosh:
Added all points listed.

PR #176, opened 2022-05-06
Format README

ivyostosh:
Update the format and language of README to improve readability

*Issue #, if available:* https://github.com/aws/eks-anywhere-packages/issues/163

*Description of changes:* `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,18 +14,18 @@ The EKS Anywhere package controller and the EKS Anywhere Curated Packages (refer
 
 ---
 
-EKS Anywhere Curated Packages is a management system for installation, configuration and maintenance of additional components for your Kubernetes cluster. Examples of these components may include Container Registry, Ingress, and LoadBalancer, etc.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,18 +14,18 @@ The EKS Anywhere package controller and the EKS Anywhere Curated Packages (refer
 
 ---
```
TerryHowe, line 16: I'm not a big fan of "The" here. It works for "The Batman", but not "The Terry Howe"

ivyostosh: I agree. This is the original approved verbiage though, so I don't want to mess up with it. I can update the rest of the README file to not use "the" convention. Does that work?

TerryHowe: Leave the disclaimer as it is, but we don't need the "the".

ivyostosh: Updated all the references except the disclaim itself.

TerryHowe (approved):
/lgtm

PR #178, opened 2022-05-06
Release bug fixes

jonahjon:
*Issue #, if available:*

*Description of changes:*

Hit an issue where the correct SHA would exist in the destination repo for promotion, but there would be a tag mismatch, causing the release job to fail. This will synchronize the tags, in addition to checking for SHA existence during the initial promotion from ECR Private to ECR Public.


generatebundlefile/ecr.go
```diff
@@ -296,6 +296,32 @@ func (c *ecrPublicClient) shaExistsInRepository(repository, sha string) (bool, e
...
+		return false, fmt.Errorf("Emtpy repository, or tag passed to the function")
+	}
+	var imagelookup []ecrpublictypes.ImageIdentifier
+	imagelookup = append(imagelookup, ecrpublictypes.ImageIdentifier{ImageTag: &tag})
+	ImageDetails, err := c.DescribePublic(&ecrpublic.DescribeImagesInput{
+		RepositoryName: aws.String(repository),
+		ImageIds:       imagelookup,
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "does not exist within the repository") == true {
```
ewollesen, line 311: In general, matching errors based on strings like this is a no-no. After the release, let's fix this up.

generatebundlefile/promote.go
```diff
@@ -120,29 +119,38 @@ func (c *SDKClients) PromoteHelmChart(repository, authFile string, crossAccount
 	if crossAccount {
 		destination = c.ecrPublicClientRelease
 	}
-	fmt.Printf("helmRequires.Spec.Images=%v\n", helmRequires.Spec.Images)
 	// Loop through each image, and the helm chart itself and check for existance in ECR Public, skip if we find the SHA already exists in destination.
 	// If we don't find the SHA in public, we lookup the tag from Private, and copy from private to Public with the same tag.
 	for _, images := range helmRequires.Spec.Images {
-		check, err := destination.shaExistsInRepository(images.Repository, images.Digest)
+		checkSha, err := destination.shaExistsInRepository(images.Repository, images.Digest)
```
ewollesen, line 125: Check your error please.

generatebundlefile/promote.go
```diff
@@ -120,29 +119,38 @@ func (c *SDKClients) PromoteHelmChart(repository, authFile string, crossAccount
...
+		checkTag, err := destination.tagExistsInRepository(images.Repository, version)
 		if err != nil {
 			return fmt.Errorf("Unable to complete sha lookup this is due to an ECRPublic DescribeImages failure %s", err)
 		}
-		if check {
-			BundleLog.Info("Image Digest already exists in destination location......skipping", images.Repository, images.Digest)
+		if checkSha && checkTag {
```
ewollesen, line 130: What if only one or the other exists? Is it okay to execute the `else` branch in that case?

ewollesen (changes requested):
The error check should be changed. The rest can wait if you want to put them off.

generatebundlefile/main.go
```diff
@@ -135,7 +135,15 @@ func main() {
 			BundleLog.Info("Starting release process....")
 			clients, err := GetSDKClients(o.release)
```
ewollesen, line 136: Heh, here's another.

generatebundlefile/promote.go
ewollesen: Jonah caught me on slack to say:
> The copy is safe, and should run if only one check passes otherwise there will be helm container pull errors.

ewollesen (changes requested):

ewollesen (commented):
/lgtm

ewollesen (approved):

PR #180, opened 2022-05-06
Update examples with latest from dev build

TerryHowe:
With these updated examples, the installs are working for me

```
% k get packages,packagebundles,packagebundlecontrollers -A
NAMESPACE       NAME                                         PACKAGE   AGE   STATE       CURRENTVERSION                                   TARGETVERSION                                             DETAIL
eksa-packages   package.packages.eks.amazonaws.com/my-test   test      8s    installed   0.1.1-6d0e566362d03a00f41d75524d21de5ddb95fd1d   0.1.1-6d0e566362d03a00f41d75524d21de5ddb95fd1d (latest)   

NAMESPACE       NAME                                                  STATE
eksa-packages   packagebundle.packages.eks.amazonaws.com/v1-21-1001   active

NAMESPACE       NAME                                                                                 STATE          DETAIL
eksa-packages   packagebundlecontroller.packages.eks.amazonaws.com/eksa-packages-bundle-controller   disconnected  
```

### DIFF
--- a/api/testdata/test.yaml
+++ b/api/testdata/test.yaml
@@ -4,7 +4,7 @@ metadata:
   name: my-test
   namespace: eksa-packages
 spec:
-  packageName: Test
+  packageName: test
   config: |
     config:
       foo: windom
--- a/api/testdata/bundle_one.yaml
+++ b/api/testdata/bundle_one.yaml
@@ -6,21 +6,21 @@ metadata:
   annotations: {}
 spec:
   packages:
-    - name: Test
+    - name: test
       source:
         registry: public.ecr.aws/l0g8r8j6
         repository: hello-eks-anywhere
         versions:
           - name: 0.1.1-6d0e566362d03a00f41d75524d21de5ddb95fd1d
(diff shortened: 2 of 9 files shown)


api/testdata/test.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: my-test
   namespace: eksa-packages
 spec:
-  packageName: Test
+  packageName: test
```
ewollesen, line 7: Did you test this? Why do I feel like this has to match line 9 exactly (ie case-sensitively)?

TerryHowe: Yeh, I tested, the comparison is case insensitive line 9 should technically be lower case too

ewollesen: 🤷🏻 ok If either works, I don't see why we're changing things, but if it works, I can't object.

/lgtm

ewollesen (commented):
/lgtm

lewisdiamond (approved):

PR #189, opened 2022-05-10
prod tag fix

jonahjon:

*Issue #, if available:*

*Description of changes:*

Tags not sync'd when copying over to Prod, trying a fix which doesn't require a new build. Otherwise we need to revert the bundle in the helm chart to `v1-22-4` since `v1-22-5` is showing newer sha sums which are not in Prod yet.


ewollesen (approved):
/lgtm
/approve

Issue #138, opened 2022-04-28
Bundle controller should ignore any bundle with wrong k8s version


TerryHowe:
Bundle controller should ignore any bundle with wrong k8s version

- [x] add method to api/v1alpha1/bundle.go something like `matchesKubeVersion(kubeVersion) bool` the returns true of the version of the bundle matches the specified kubernetes version
- [x] Add some new state to the bundle state like `WRONGK8S`, that is a terrible name, come up with something better
- [x] the bundler controller reconciler should mark the bundle as `WRONGK8S` if the current cluster kubeVersion doesn't match the bundle

PR #203, opened 2022-05-12
Add KubeVersion check to PackageBundle State

ivyostosh:
If the version of the bundle doesn't match the specified Kubernetes version, the PackageBundle state will be marked as `PackageBundleStateIgnored`.

*Issue #, if available:* https://github.com/aws/eks-anywhere-packages/issues/138

*Description of changes:* Add `KubeVersionMatches` method, helper function `GetMajorMinorFromString`; and update `Update` in manager.go

### DIFF
--- a/api/v1alpha1/packagebundle.go
+++ b/api/v1alpha1/packagebundle.go
@@ -70,7 +70,7 @@ func (config *PackageBundle) getMajorMinorBuild() (major int, minor int, build i
 }
 
 // getMajorMinorFromString returns the Kubernetes major and minor version.
-
+//
 // It returns 0, 0 for empty string.
 func getMajorMinorFromString(kubeVersion string) (major int, minor int) {
 	s := strings.Split(kubeVersion, "-")
--- a/api/v1alpha1/packagebundle.go
+++ b/api/v1alpha1/packagebundle.go
@@ -57,8 +57,8 @@ func (config *PackageBundle) LessThan(rhsBundle *PackageBundle) bool {
 	return lhsMajor < rhsMajor || lhsMinor < rhsMinor || lhsBuild < rhsBuild
 }
 
-// getMajorMinorBuild parses PackageBundle name, and return the Kubernetes major
-// version, Kubernetes minor version, and bundle build version.
+// getMajorMinorBuild returns the Kubernetes major version, Kubernetes minor
+// version, and bundle build version.
 func (config *PackageBundle) getMajorMinorBuild() (major int, minor int, build int) {
 	s := strings.Split(config.Name, "-")
 	s = append(s, "", "", "")
@@ -69,8 +69,7 @@ func (config *PackageBundle) getMajorMinorBuild() (major int, minor int, build i
 	return major, minor, build
 }
 
-// getMajorMinorFromString parses the kubeVersion (in string format), and
-// return the Kubernetes major version and Kubernetes minor version.
+// getMajorMinorFromString returns the Kubernetes major and minor version.
 
 // It returns 0, 0 for empty string.
 func getMajorMinorFromString(kubeVersion string) (major int, minor int) {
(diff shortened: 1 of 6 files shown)


pkg/bundle/manager.go
```diff
@@ -19,7 +19,7 @@ import (
 
 type Manager interface {
 	// Update the bundle returns true if there are changes
```
TerryHowe, line 21: The pattern where we pass in the state to `Update` is opposite of the pattern I have been trying to follow here. I've been trying to get around to where `isActive` is not passed in here. It has been difficult getting rid of that and this adds another state we are passing.

api/v1alpha1/packagebundle.go
```diff
@@ -67,12 +69,34 @@ func (config *PackageBundle) GetMajorMinorBuild() (major int, minor int, build i
 	return major, minor, build
 }
```
TerryHowe, line 71: It looks like this could be a private method and maybe the other method could call it? I'm okay testing private methods

TerryHowe (changes requested):

controllers/packagebundle_controller.go
```diff
@@ -124,7 +124,7 @@ func (r *PackageBundleReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 			return ctrl.Result{}, nil
 		}
 
-		// Downlod the bundle using name tag.
+		// Download the bundle using name tag.
```
ewollesen, line 127: 👍🏻

api/v1alpha1/packagebundle_test.go
```diff
@@ -106,7 +106,87 @@ func TestPackageBundle_Find(t *testing.T) {
 	})
 }
 
-func TestMatches(t *testing.T) {
+func TestGetMajorMinorFromString(t *testing.T) {
+
+	t.Run("Parse from default Kubernetes version name", func(t *testing.T) {
+
+		target_version := "v1-21-1"
```
ewollesen, line 113: nit: [Golang convention](https://go.dev/doc/effective_go#mixed-caps) is to use targetVersion rather than target_version.

api/v1alpha1/packagebundle.go
```diff
@@ -67,12 +69,34 @@ func (config *PackageBundle) GetMajorMinorBuild() (major int, minor int, build i
...
 
+// GetMajorMinorFromString parses the kubeVersion (in string format), and
+// return the Kubernetes major version and Kubernetes minor version. It returns
+// 0, 0 for empty string.
+func GetMajorMinorFromString(kubeVersion string) (major int, minor int) {
```
ewollesen, line 75: nit: From _Effective Go_:
>  The first sentence should be a one-sentence summary that starts with the name being declared. 
-- https://go.dev/doc/effective_go#commentary

ewollesen (commented):
Nice work. I just pointed out a few nit picks about documentation style. Take them or leave them.

Regarding the overall PR though, where / how do you forsee KubeVersionMatches being used?

pkg/bundle/manager.go
ivyostosh: Fixed in the latest commit

api/v1alpha1/packagebundle.go
ivyostosh: I made the method to private and created an export file to finish unit testing (open to other suggestions).

api/v1alpha1/packagebundle_test.go
ivyostosh: Fixed

api/v1alpha1/packagebundle.go
ivyostosh: Updated

TerryHowe (approved):

TerryHowe (approved):
/lgtm

PR #210, opened 2022-05-19
Don't leak credentials via process table inspection.

ewollesen:
When credentials are part of a command's arguments, like these were,
they're exposed via process table inspection (e.g. via `ps`). In
addition, they could possibly then be captured in logs.

The replacement defines a function that wraps the oras call and passes
in the credentials on stdin, removing them from the process table and
preventing exposure via logs.

*Issue #, if available:*

*Description of changes:*


TerryHowe (approved):
/lgtm
/approve

TerryHowe (approved):
/lgtm
/approve

PR #212, opened 2022-05-19
Ensure that the output directory exists.

ewollesen:
The creation of the directory is being performed in scripts, but they
shouldn't need to worry about things like this. Move it to the
Makefile where it belongs.

*Issue #, if available:*

*Description of changes:*


TerryHowe (approved):
/lgtm
/approve

PR #213, opened 2022-05-19
Extract functions and use a loop to be DRYer.

ewollesen:
This script was doing the exact same things for two different k8s
versions. I extracted that behavior into two orthogonal functions,
then used a loop to remove the duplication.

Future improvements would be to modify the generatebundle binary so
that its output is more predictable, and make the versions in the
input files match the values used here.

Edit: As @TerryHowe pointed out, we should also work towards a place where the files present determine the versions of k8s that we're supporting.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/generatebundlefile/hack/generate_bundle.sh
+++ b/generatebundlefile/hack/generate_bundle.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+
 # Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


ewollesen:
/hold 
Depends on 212

generatebundlefile/hack/generate_bundle.sh
```diff
@@ -13,39 +14,43 @@
...
+    local version=$1
+
+    cd "${BASE_DIRECTORY}/generatebundlefile/output"
+    awsAuth "$ORAS_BIN" push --username AWS --password-stdin \
+            "${REPO}:v${version}-${CODEBUILD_BUILD_NUMBER}" bundle.yaml
+    awsAuth "$ORAS_BIN" push --username AWS --password-stdin \
+            "${REPO}:v${version}-latest" bundle.yaml
+}
+
+for version in 1-21 1-22; do
```
TerryHowe, line 53: at some point we the k8s versions based on the existence of the various input yaml files

TerryHowe (approved):
/lgtm
/approve

generatebundlefile/hack/generate_bundle.sh
ewollesen: Oh, yeah, that'd be good too. We should get the files named better so we don't have to tweak things as much too.

TerryHowe (approved):
/lgtm
/approve

PR #215, opened 2022-05-21
Get image tags from files

TerryHowe:
Pull the image tags from files for now so we can modify them easily


jonahjon:
Works all good

TerryHowe:
> @TerryHowe where do you see us pulling them from in the future?

I'm not sure, but this will work well if we are going to put these in manually or using a bot. More deterministic than "latest" at least, people will know what they are getting.

jonahjon (approved):

Issue #204, opened 2022-05-12
When controller downloads new bundle, state is not set


TerryHowe:
Fresh install of the package controller and it downloaded a new bundle and did not set the state to inactive
```
% k get packages,packagebundles,packagebundlecontrollers -A
NAMESPACE       NAME                                               STATE
eksa-packages   packagebundle.packages.eks.amazonaws.com/v1-21-8   
eksa-packages   packagebundle.packages.eks.amazonaws.com/v1-22-5   active (upgrade available)
```

TerryHowe:
some error messages in log, not sure if they matter, maybe the webhook just hadn't started yet
```
1.6523846470133932e+09	ERROR	controller.packagebundlecontroller	Reconciler error	{"reconciler group": "packages.eks.amazonaws.com", "reconciler kind": "PackageBundleController", "name": "eksa-packages-bundle-controller", "namespace": "eksa-packages", "error": "creating new package bundle: Internal error occurred: failed calling webhook \"vpackagebundle.kb.io\": Post \"https://eks-anywhere-packages-eks-anywhere-packages-webhook-service.eksa-packages.svc:443/validate-packages-eks-amazonaws-com-v1alpha1-packagebundle?timeout=10s\": dial tcp 10.110.224.24:443: connect: connection refused"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:227
1.6523846474863873e+09	ERROR	controller.packagebundlecontroller	Reconciler error	{"reconciler group": "packages.eks.amazonaws.com", "reconciler kind": "PackageBundleController", "name": "eksa-packages-bundle-controller", "namespace": "eksa-packages", "error": "creating new package bundle: Internal error occurred: failed calling webhook \"vpackagebundle.kb.io\": Post \"https://eks-anywhere-packages-eks-anywhere-packages-webhook-service.eksa-packages.svc:443/validate-packages-eks-amazonaws-com-v1alpha1-packagebundle?timeout=10s\": dial tcp 10.110.224.24:443: connect: connection refused"}
[...]

PR #217, opened 2022-05-24
Remove default bundle from chart

TerryHowe:
Closes: https://github.com/aws/eks-anywhere-packages/issues/204

### DIFF
--- a/controllers/packagebundlecontroller_controller.go
+++ b/controllers/packagebundlecontroller_controller.go
@@ -115,10 +115,11 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 		return withoutRequeue(result), nil
 	}
 
+	result.RequeueAfter = pbc.Spec.UpgradeCheckShortInterval.Duration
 	latestBundle, err := r.bundleManager.LatestBundle(ctx, pbc.Spec.Source.BaseRef())
 	if err != nil {
 		r.Log.Error(err, "Unable to get latest bundle")
-		if pbc.Status.State != api.BundleControllerStateDisconnected {
+		if pbc.Status.State == api.BundleControllerStateActive {
 			pbc.Status.State = api.BundleControllerStateDisconnected
 			r.Log.V(6).Info("update", "PackageBundleController", pbc.Name, "state", pbc.Status.State)
 			err = r.Client.Status().Update(ctx, pbc, &client.UpdateOptions{})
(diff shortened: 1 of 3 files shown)


controllers/packagebundlecontroller_controller.go
```diff
@@ -147,12 +146,22 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
...
 	}
 
+	if pbc.Spec.ActiveBundle == "" {
+		r.Log.V(6).Info("Activating new bundle", "bundle", latestBundle.Name)
+		pbc.Spec.ActiveBundle = latestBundle.Name
+		err = r.Client.Update(ctx, pbc)
+		if err != nil {
+			return result, fmt.Errorf("updating active bundle: %s", err)
+		}
+		return result, fmt.Errorf("updating active bundle: %s", err)
```
a-cool-train, line 163: is this a typo? Don't we want to return nil in case the update is successful?

controllers/packagebundlecontroller_controller.go
```diff
@@ -147,12 +146,22 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
...
 
 	if !r.bundleManager.IsBundleKnown(ctx, bundles.Items, latestBundle) {
+		r.Log.V(6).Info("New latest bundle", "bundle", latestBundle.Name)
 		err = r.Client.Create(ctx, latestBundle)
 		if err != nil {
 			return result, fmt.Errorf("creating new package bundle: %s", err)
 		}
 	}
 
+	if pbc.Spec.ActiveBundle == "" {
```
ewollesen, line 156: I have concerns about the implications of this flow. It's not clear to me that when there is no active bundle, that the right choice is to activate the newest bundle.

Outside of first-run conditions, this should never happen, but it if does, it could be disastrous to upgrade someone's bundle out from under them. There's no logic here to check for a first-run condition (which is why I like having a bundle from the start, it eliminates the need for a first-run check, and makes any case where activebundle is blank an error state).

I think I'd feel a lot better if there were something else in the spec or status of a new controller that indicated, through it's presence (rather than through its absence) that this was a first-run situation.

controllers/packagebundlecontroller_controller.go
TerryHowe: Yeh, fixed thanks

controllers/packagebundlecontroller_controller.go
TerryHowe: Let me think what we can do here.

TerryHowe:
/hold
Going to try another PR with some other changes to work to something more reasoable

controllers/packagebundlecontroller_controller.go
```diff
@@ -115,35 +115,48 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 		return withoutRequeue(result), nil
 	}
 
+	result.RequeueAfter = pbc.Spec.UpgradeCheckShortInterval.Duration
 	latestBundle, err := r.bundleManager.LatestBundle(ctx, pbc.Spec.Source.BaseRef())
 	if err != nil {
 		r.Log.Error(err, "Unable to get latest bundle")
-		if pbc.Status.State != api.BundleControllerStateDisconnected {
+		if pbc.Status.State == api.BundleControllerStateActive {
```
vincentni, line 122: Why do we want to change this condition? I'm aware that currently there are three possible states for bundle controller, and BundleControllerStateIgnored state won't reach here as it gets returned right before. But won't the original condition be much safer for future extension with more states added?

TerryHowe: Because we are now supporting a brand new PBC which would have the state `""`.  I do not want to change the state to disconnected if we haven't activated a bundle yet.

vincentni (approved):

Issue #164, opened 2022-05-05
Increase code coverage of pkg/bundle


TerryHowe:
Our test code coverage is 50% for pkg/bundle which is not great for some pretty key code. It should be closer to 80%

PR #219, opened 2022-05-25
Adding code coverage for pkg/bundle

junshun:
Closes: #164 

*Description of changes:*
Unit testing was added for IsBundleKnown and LatestBundle for the bundle manager.

Unit Test was created for client with mock interfaces added

Coverage is now at 92% for pkg/bundle. Most of the remaining percentage is in the mock code for client.

### DIFF
--- a/pkg/bundle/client_test.go
+++ b/pkg/bundle/client_test.go
@@ -5,13 +5,12 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+	ctrlmocks "github.com/aws/eks-anywhere-packages/controllers/mocks"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/bundle/client_test.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+}
+
+// Helpers
+type mockClient struct {
+	client.Client
+	err error
+}
+
+func newMockClient(err error) *mockClient {
```
TerryHowe, line 133: Can this be thrown in this https://github.com/aws/eks-anywhere-packages/blob/main/Makefile#L163

junshun: Addressed in latest commit. Used the mock client interface that was already generated in controllers/mocks

pkg/bundle/client_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+		namespacedNames, err := bundleClient.GetActiveBundleNamespacedName(ctx)
+
+		assert.Equal(t, api.PackageNamespace, namespacedNames.Namespace)
+		assert.Equal(t, "", namespacedNames.Name)
+		assert.Nil(t, err)
+	})
+}
+
+// Helpers
```
TerryHowe, line 138: My minor nit on these helper names, for tests, I normally call these something like `givenMockClient` so they don't show up when someone is trying to remember what the name of some new function. Also given follows the testing given/when/then pattern. We are putting tests in the same package as our code and the danger with doing that is someone will accidentally use test code in a production implementation.

TerryHowe (approved):
/lgtm
/approve

Issue #216, opened 2022-05-23
Investigate if we can Remove Kube-RBAC-Proxy, and remove if possible.


jonahjon:
**What would you like to be added**:

**Why is this needed**:

PR #225, opened 2022-06-02
Removing kube rbac proxy from helm chart

jonahjon:

*Issue #, if available:*
https://github.com/aws/eks-anywhere-packages/issues/216

*Description of changes:*
Remove Kube RBAC proxy from helm chart, since we don't need/use it properly.


TerryHowe (approved):
/lgtm
/approve

Issue #20, opened 2022-03-02
Allow registry credentials to be passed to the helm driver


lewisdiamond:
The helm driver currently only works on public registries. Add a configuration option that allows to use a secret with credentials for pulling artifacts with the helm registry client.

jonahjon:
Did did this for the build infrastructure using an Authfile to allow private/public within the client.

PR #251, opened 2022-06-11
Passing Registry Credentials to Helm Driver

junshun:
Closes issue #20 

*Description of changes:*
Adding the ability to pass in a docker secret as an environment variable through the helm chart for the package controller. With this change the helm driver will be able to pull hem charts from a private repository. Changes to the controller helm chart will be seperate from this commit.


pkg/driver/helmdriver.go
```diff
@@ -30,8 +31,13 @@ var _ PackageDriver = (*helmDriver)(nil)
 
 func NewHelm(log logr.Logger) (*helmDriver, error) {
 	settings := cli.New()
-	//TODO: Allow configuring the client's credentials from a secret
-	//see: https://github.com/aws/eks-anywhere-packages/issues/20
+
+	// TODO Catch error here if not provided docker config or continue without an authfile if possible
+	secretAuth := auth.NewDockerSecret()
```
TerryHowe, line 36: I would imagine Eric is going to ask you to inject this

TerryHowe (changes requested):
Generally looks good except the conflicting use of DOCKER_CONFIG

pkg/driver/helmdriver.go
junshun: This will be handled in a future issue along with updating the helm driver should the provided credentials be updated ( which could happen if login credentials time out for example)

pkg/authenticator/dockersecret.go
```diff
@@ -0,0 +1,45 @@
...
+	token, err := getSecretToken()
+	if err != nil {
+		return "", fmt.Errorf("Failed to get authfile %s\n", err)
+	}
+	authfile, err := createAuthFile(token)
+	if err != nil {
+		return "", fmt.Errorf("Failed to get authfile %s\n", err)
+	}
+
+	return authfile, nil
```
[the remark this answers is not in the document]
junshun: Are you suggesting adding a volume mount which is used for just holding registry credentials?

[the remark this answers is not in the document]
junshun: So I had this implemented in a branch here https://github.com/junshun/eks-anywhere-packages/tree/authn.

That was my original implementation, for it I had to change the NewHelm to take in mgr.config so then I can get the secret. The only reason I moved to this environment variable solution was it was easier to local testing and didn't require the rest.config to be passed into the helm driver. 

Currently my idea for specifing this enviroment variable is via the helm chart in the deployment.yaml so it would be a docker formated mounted secret but to the package controller pod it shows up as an environment variable
```
          env:
            - name: "DOCKER_CONFIG"
              valueFrom:
                secretKeyRef:
                  name: dockerconfig
                  key: .dockerconfigjson
```

pkg/driver/helmdriver.go
```diff
@@ -30,8 +31,13 @@ var _ PackageDriver = (*helmDriver)(nil)
 
 func NewHelm(log logr.Logger) (*helmDriver, error) {
 	settings := cli.New()
-	//TODO: Allow configuring the client's credentials from a secret
-	//see: https://github.com/aws/eks-anywhere-packages/issues/20
+
+	// TODO Catch error here if not provided docker config or continue without an authfile if possible
+	secretAuth := auth.NewHelmSecret()
+	authfile, _ := secretAuth.GetAuthFileName()
+	if authfile != "" {
```
TerryHowe, line 38:
```suggestion
	if len(authfile) != 0 {
```

len handles nil too

pkg/authenticator/helmsecret.go
```diff
@@ -0,0 +1,33 @@
...
+
+func (s *helmsecret) GetAuthFileName() (string, error) {
+	// Check if Helm registry config is set
+	helmconfig := os.Getenv("HELM_REGISTRY_CONFIG")
+	if helmconfig != "" {
+		// Use HELM_REGISTRY_CONFIG
+		return helmconfig, nil
+	}
+
+	// Check secret mount
```
TerryHowe, line 24: How about it just returns nil if HELM_REGISTRY_CONFIG is not set. This /secrets is a pod things, I'm not entirely sure it is correct and gosec doesn't like this.

TerryHowe (approved):

pkg/authenticator/helmsecret.go
junshun: Well we need to pass in this authfile for the helm driver if it does exist. When this is running in a pod it would be there if its set and then it should be pass to the helm driver.

TerryHowe: The pod should set HELM_REGISTRY_CONFIG

TerryHowe: s/pod/deployment in the helm chart. The thing that sets where it is should set where it can be found.

docs/README.md
```diff
@@ -17,23 +17,27 @@ Create a cluster with EKS Anywhere and set and export with KUBECONFIG.
 
         make install
 
-1. Run the controller locally:
+2. Run the controller locally:
```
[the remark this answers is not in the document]
ewollesen: It will be numbered automatically when it's rendered, but the source needn't be for that to happen. A lot of editors have a plugin for markdown that will do the renumbering, but that's a separate (and technically unneeded) thing.

pkg/authenticator/authenticator.go
```diff
@@ -0,0 +1,19 @@
+package authenticator
```
ewollesen, line 1: Minor thing, but this filename isn't great, as it's too generic. What is it authenticating to?

docs/README.md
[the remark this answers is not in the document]
junshun: Yep my editor change this on its own. I ended up editing it in vim so should be good now

pkg/authenticator/authenticator.go
```diff
@@ -0,0 +1,19 @@
...
+	Auth string `json:"auth"`
+}
+
+// Authenticator is an interface for creating an authentication file with credentials to private registries
+//
+// Currently this is used with the Helm Driver which takes credentials in this way
+// For this first implementation, kubernetes secrets will be used to pass in a token
+type Authenticator interface {
+	// GetAuthFileName Gets Authentication File Path for OCI Registry
+	GetAuthFileName() (string, error)
```
ewollesen, line 18:
```suggestion
	AuthFilename() (string, error)
```
Nit: There's no need for the `Get` prefix, and in go, it's usually capitalized as `Filename` not `FileName`.

pkg/authenticator/helmsecret.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"os"
+)
+
+type helmsecret struct {
+}
+
+var _ Authenticator = (*helmsecret)(nil)
+
+func NewHelmSecret() *helmsecret {
```
ewollesen, line 12: Nit: Try to be consistent with your capitalization. That is, pick one of either helmSecret or helmsecret, but not both. I prefer helmSecret, because while I'm all for a good [compound word](https://en.wikipedia.org/wiki/English_compound), I don't think helmsecret qualifies. :)

pkg/authenticator/helmsecret.go
```diff
@@ -0,0 +1,33 @@
...
+	return &helmsecret{}
+}
+
+func (s *helmsecret) GetAuthFileName() (string, error) {
```
ewollesen, line 16:
```suggestion
func (s *helmsecret) AuthFilename() (string, error) {
```
If you want to match a previous nit.

pkg/authenticator/helmsecret.go
ewollesen: I agree with Terry here. This code should know as little as possible about how k8s works, and the fact that secrets are usually put in some specific place is 100% a k8s concern. This code should just look for the auth where the env var tells it.

pkg/authenticator/helmsecret_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewHelmSecret(t *testing.T) {
+	t.Parallel()
+
+	t.Run("golden path", func(t *testing.T) {
```
ewollesen, line 13: There's really no need for this test. It's not even a "golden path", because it doesn't actually _do anything_. While I can see the usefulness of such a test early on as you're writing the code, it's not worth making every dev run this test every time they run `make test`.

pkg/driver/helmdriver.go
TerryHowe: Disregard this.

pkg/authenticator/helmsecret_test.go
junshun: Good call, this actually did more in the original implementation which is why it got stuck here.

TerryHowe (approved):
/lgtm
/approve

PR #252, opened 2022-06-11
Add UpdateLatestBundle method to manager

TerryHowe:
Move some code from reconciler to bundle manager


controllers/packagebundlecontroller_controller.go
```diff
@@ -140,14 +140,8 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 		}
 	}
 
-	bundles := &api.PackageBundleList{}
-	err = r.Client.List(ctx, bundles, &client.ListOptions{Namespace: api.PackageNamespace})
+	err = r.bundleManager.UpdateLatestBundle(ctx, latestBundle)
```
ewollesen, line 143: I imagine your next move might be to combine this call with the `LatestBundle` call above?

ewollesen (commented):
/lgtm

controllers/packagebundlecontroller_controller.go
TerryHowe: Yeh, definitely.

PR #255, opened 2022-06-11
Make registry optional in bundle

TerryHowe:
The registry in the bundle should not be provided unless we want to override it. This change is to support private registries.


lewisdiamond:
Does this work? Seems like with this change, not providing a registry in the bundle will lead to failure to install anything. The registry isn't currently defaulted to the controller's value.

Could you also add tests for cases where it's provided (i.e. gets overridden) and not provided (i.e. gets defaulted)?

lewisdiamond:
/hold
This depends on #256

TerryHowe:
>  This depends on https://github.com/aws/eks-anywhere-packages/pull/256

Yes

Issue #237, opened 2022-06-03
Ignore bundles outside of namespace


TerryHowe:
Anything outside of eksa-packages should be ignored and marked ignored

PR #257, opened 2022-06-15
Use get bundle list in manager

TerryHowe:
Use of bundle list in controller was not filtering by namespace and this change uses a common method to get the list.

Slight added benefit that we don't get the bundle list unless we need to.

Closes: #237


ewollesen (commented):
/lgtm

This is a great change

PR #259, opened 2022-06-16
Rename UpdateLatestBundle to ProcessLatestBundle

TerryHowe:
This method doesn't update, it creates if needed.


ewollesen (commented):
/lgtm

PR #265, opened 2022-06-16
Fixing pipeline removing registry from bundle output

jonahjon:

*Description of changes:*

We keep the registry in the PackageBundleSpec until after the lookups are complete, then set it to empty string before doing the signature, and writing to yaml.


TerryHowe (approved):
/lgtm
/approve

PR #266, opened 2022-06-16
Removing signature flag not used anymore

jonahjon:

*Description of changes:*

This was used when we were using Cosign to sign, but not anymore, cleaning up the code a bit with this removal.


TerryHowe (approved):
/lgtm
/approve

PR #267, opened 2022-06-17
Temp-fix for config types in marshaling

jonahjon:

*Issue #, if available:*

*Description of changes:*

Config Values for Helm charts come in multiple different types, and it's breaking the marshaling trying since things are currently only done as strings. This is a temp fix until we figure out a better long term solution for config checking.

https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/goharbor/harbor/helm/requires-config.yaml#L26

Created an issue for follow-up on this with a better long term solution. https://github.com/aws/eks-anywhere-packages/issues/270

### DIFF
--- a/generatebundlefile/bundle_types.go
+++ b/generatebundlefile/bundle_types.go
@@ -96,9 +96,9 @@ type RequiresSpec struct {
 }
 
 type Configuration struct {
-	Name     string `json:"name,omitempty"`
-	Required bool   `json:"required,omitempty"`
-	Default  string `json:"default,omitempty"`
+	Name     string      `json:"name,omitempty"`
+	Required bool        `json:"required,omitempty"`
+	Default  interface{} `json:"default,omitempty"`
 }
 
 type Image struct {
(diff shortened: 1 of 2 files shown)


generatebundlefile/bundle_types.go
```diff
@@ -96,9 +96,9 @@ type RequiresSpec struct {
 }
 
 type Configuration struct {
-	Name     string `json:"name,omitempty"`
-	Required bool   `json:"required,omitempty"`
-	Default  string `json:"default,omitempty"`
+	Name     string      `json:"name,omitempty"`
+	Required bool        `json:"required,omitempty"`
+	Default  interface{} `json:"default,omitempty"`
```
TerryHowe, line 101: What kind of defaults are giving you problems? Can it be treated like embedded yaml? This interface{} is probably not going to work

TerryHowe: That `80` doesn't work as a string?

jonahjon: There is int and bool in this. https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/goharbor/harbor/helm/requires-config.yaml#L26

jonahjon: those are giving the errors

TerryHowe: Put quotes around them

jonahjon:
Closing this

PR #268, opened 2022-06-17
Updating STS client connection

jonahjon:

*Issue #, if available:*

*Description of changes:*

Changing the sts client connection to more modular for upcoming PR in which we need to create connection with profile to another account.


TerryHowe (approved):
/lgtm

PR #271, opened 2022-06-20
Fix to keep helm value types

jonahjon:

*Issue #, if available:*

*Description of changes:*

This is a work-around that's from this thread, and comment https://github.com/kubernetes-sigs/controller-tools/pull/126#issuecomment-630764976

and this thread in slack https://kubernetes.slack.com/archives/CAR30FCJZ/p1602513628452900

Tested out on presubmit, and generated a bundle that had support for multiple types from the requires.yaml config.

These types are supported: bool, int64, float64, string, []interface{}, map[string]interface{} and nil, which should support almost all of our use cases too.


api/v1alpha1/packagebundle_types.go
```diff
@@ -104,7 +105,8 @@ type VersionConfiguration struct {
 	Required bool `json:"required"`
 
 	// Default is the name of the configuration
-	Default string `json:"default"`
+	// +kubebuilder:validation:Optional
+	Default *apiextensions.JSON `json:"default,omitempty"`
```
TerryHowe, line 109: This looks real interesting although I'm a bit concerned about not having much time to mess with this before our release (which in theory should be out already)

jonahjon:
Closing since we are in favor of schemas

PR #292, opened 2022-06-28
Fix some logs and spelling

TerryHowe:
Fix some logs and spelling errors

### DIFF
--- a/pkg/bundle/manager.go
+++ b/pkg/bundle/manager.go
@@ -81,6 +81,7 @@ func (m bundleManager) Update(ctx context.Context, newBundle *api.PackageBundle)
 	}
 	if !active {
 		if newBundle.Status.State == api.PackageBundleStateActive {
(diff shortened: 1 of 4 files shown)


pkg/bundle/manager.go
```diff
@@ -81,6 +81,7 @@ func (m bundleManager) Update(ctx context.Context, newBundle *api.PackageBundle)
 	}
 	if !active {
 		if newBundle.Status.State == api.PackageBundleStateActive {
+			m.log.V(6).Info("mark bundle inactive", "bundle", newBundle.Name)
```
ivyostosh, line 84: This comment applies to all logs in this file.

Will it be better to have the log after setting the newBundle.Status.State? This is to avoid the case where users think their state has been set by looking at this log, but somehow the Status.State didn't get assigned properly. This can be a misleading log IMO.

TerryHowe: Done. (I think)

PR #294, opened 2022-06-28
Remove PackageController CRD

TerryHowe:
We haven't used this CRD yet, I doubt we will any time soon


ewollesen:
As discussed previously, I believe these were intended for persistence of packages in the face of external manipulation. That is, in the event someone runs `helm uninstall` on a package managed by our packages controller, it would be the job of the PackageController (now removed) to re-install that helm chart as part of its reconcile loop. As Terry points out, haven't yet implemented that behavior, and this code has just been sitting there ever since. By getting rid of it, we reduce some cruft for now. Hopefully we won't forget about this when it's time to introduce this behavior.

/lgtm

ewollesen (commented):
/lgtm
/approve

ewollesen:
Am I not allowed to do two comments at once? Or was the bot just down for an oil change/

/lgtm

PR #295, opened 2022-06-29
Activate PBC first

TerryHowe:
Activate the PBC first and then on another pass, get the latest bundle.

### DIFF
--- a/controllers/packagebundlecontroller_controller.go
+++ b/controllers/packagebundlecontroller_controller.go
@@ -104,8 +104,8 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 
 	if pbc.IsIgnored() {
 		if pbc.Status.State != api.BundleControllerStateIgnored {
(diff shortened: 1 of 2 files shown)


controllers/packagebundlecontroller_controller.go
```diff
@@ -114,12 +114,22 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
...
+		if err != nil {
+			return result, fmt.Errorf("updating status: %s", err)
+		}
+		return result, nil
+	}
+
 	latestBundle, err := r.bundleManager.LatestBundle(ctx, pbc.Spec.Source.BaseRef())
 	if err != nil {
 		r.Log.Error(err, "Unable to get latest bundle")
 		if pbc.Status.State != api.BundleControllerStateDisconnected {
```
ivyostosh, line 130: Is this check still necessary? Shouldn't pbc.Status.State be active for all pbc that can reach this step?

PR #298, opened 2022-06-29
Add ignored version state to bundle

TerryHowe:


api/v1alpha1/packagebundle_types.go
```diff
@@ -113,14 +113,15 @@ type PackageBundleStatus struct {
...
 
-//+kubebuilder:validation:Enum={"inactive","active","active (upgrade available)","ignored"}
+//+kubebuilder:validation:Enum={"inactive","active","active (upgrade available)","ignored","ignored version"}
 type PackageBundleStateEnum string
 
 const (
 	PackageBundleStateInactive         PackageBundleStateEnum = "inactive"
 	PackageBundleStateActive           PackageBundleStateEnum = "active"
 	PackageBundleStateUpgradeAvailable PackageBundleStateEnum = "active (upgrade available)"
 	PackageBundleStateIgnored          PackageBundleStateEnum = "ignored"
```
ivyostosh, line 123: I find "ignored" and "ignored version" not mutually exclusive by description. Is there a better name convention that can help address this issue?

TerryHowe: Yeh, I'm not a big fan of the name either. I don't like state names with spaces in there. I don't like "active (upgrade available)", but that is where we went for for whatever reason. I'd kind of sooner PackageBundleStateUpgradeAvailable didn't exist just PackageBundleStateActive with a NOTE "upgrade available"

ivyostosh: I hear what you are saying. Not for this PR, but I think if we are having doubt about those names, we should change it prior to GA. It seems to be better to address it early on before we accumulate a bigger user base.

PR #304, opened 2022-07-01
Reactivate the package controller

TerryHowe:
When I was removing the packagecontroller_controller.go, I must have removed this call on accident.


ewollesen (commented):
I'm surprised we still need this. The naming has never been ideal. Sigh.

/lgtm

PR #305, opened 2022-07-01
Rename Update method to ProcessBundle

TerryHowe:
This generic name was bugging me


ivyostosh:
This generic name has been bothering me for a while as well. Great to see it gets changed!

/lgtm
/approve

PR #311, opened 2022-07-19
Feature config validation

a-cool-train:
*Description of changes:*
Here are the changes contained in this PR:
- Include package validation webhook
- Parse json schema contained in a bundle: decode, then unzip
- Include Functional tests for the feature.


api/v1alpha1/package_webhook.go
```diff
@@ -0,0 +1,206 @@
...
+		return admission.Errored(http.StatusInternalServerError,
+			fmt.Errorf("decoding request: %w", err))
+	}
+
+	bundles, err := v.listBundles(ctx)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError,
+			fmt.Errorf("listing package bundles: %w", err))
+	}
+
```
TerryHowe, line 64:
```suggestion
```

TerryHowe (changes requested):

api/v1alpha1/package_webhook.go
```diff
@@ -0,0 +1,206 @@
...
+	pbc, err := v.getActiveController(ctx)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("getting PackageBundleController: %v", err))
+	}
+
+	activeBundle := v.getActiveBundle(bundles, pbc.Spec.ActiveBundle)
```
lewisdiamond, line 70: `getActiveBundle` should really be renamed or reworked to figure out which bundle is active.

api/v1alpha1/package_webhook.go
```diff
@@ -0,0 +1,206 @@
...
+	var activeBundle PackageBundle
+	for _, bundle := range bundles.Items {
+		if bundle.Name == b {
+			activeBundle = bundle
+		}
+	}
+	return activeBundle
+}
+
+func (v *packageValidator) getActiveController(ctx context.Context) (PackageBundleController, error) {
```
lewisdiamond, line 101: `getActiveController` and `getActiveBundle`/`getBundle` should be part of a more structured API for all our resources. I'd be fine if we consider this out-of-scope for this PR but we need to get started on that as it will help with testing as well.

api/v1alpha1/package_webhook.go
```diff
@@ -0,0 +1,206 @@
...
+
+	packageVersions := packageInBundle.Source.Versions
+	if len(packageVersions) < 1 {
+		return false, fmt.Errorf("package %s does not contain any versions", p.Name)
+	}
+
+	// The package configuration is gzipped and base64 encoded
+	// When processing the configuration, the reverse occurs: base64 decode, then unzip
+	configuration := packageInBundle.Source.Versions[0].Configurations[0].Default
+	decodedConfiguration, err := base64.StdEncoding.DecodeString(configuration)
```
lewisdiamond, line 131: Wrap this in the function that gets the schema instead.

api/v1alpha1/package_webhook.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error when uncompressing configurations %v", err)
+	}
+
+	output, err := ioutil.ReadAll(gzreader)
+	if err != nil {
+		return nil, fmt.Errorf("error reading configurations %v", err)
+	}
+
+	jsonSchema, err := yaml.YAMLToJSON(output)
```
lewisdiamond, line 186: Why do we need "yaml to json" here? The jsonschema is in JSON format already.

lewisdiamond:
This webhook should be defaulting + validating. We need to set the default values from what's specified in the jsonschema rather than only accepting/refusing the resource.

a-cool-train:
> This webhook should be defaulting + validating. We need to set the default values from what's specified in the jsonschema rather than only accepting/refusing the resource.

That's a great idea. Let me add the defaulting feature.

a-cool-train:
> This webhook should be defaulting + validating. We need to set the default values from what's specified in the jsonschema rather than only accepting/refusing the resource.

@lewisdiamond @TerryHowe, I was thinking about this a bit and I realized that the CLI already creates defaulting values in both install and generate packages commands. Doing this in the controller will be duplicating the work and on top of that adding extra processing time to the controller that might not be necessary.

pkg/testutil/packagevalidation.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/aws/eks-anywhere-packages/api/v1alpha1"
+)
+
```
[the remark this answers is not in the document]
a-cool-train: Ok as discussed, I removed the extra methods that duplicate the work.

api/v1alpha1/package_webhook.go
a-cool-train: There is actually an api that already exists for it but including it in my code causes an import cycle which can't be solved without a huge revamp of the codebase. We can rethink of this in a more structured manner in a later PR.

api/testdata/package_webhook_invalid_config.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: packages.eks.amazonaws.com/v1alpha1
+kind: Package
+metadata:
+  name: my-hello-eks-anywhere
+  namespace: eksa-packages
+spec:
+  packageName: hello-eks-anywhere
+  config: |
+    sourceRegistry: "public.ecr.aws/dev"
+    fakeConfig: "Amazon EKS Anywhere"
```
lewisdiamond, line 10: Configure your editor to add new lines at end of files.

api/v1alpha1/webhook_suite_test.go
```diff
@@ -73,40 +73,91 @@ func TestAPIs(t *testing.T) {
 	RunSpecs(t, "Webhook Suite", suiteConfig, reporterConfig)
 }
 
-var _ = Context("Releases signatures are validated against the correct public key", func() {
```
lewisdiamond, line 76: Can you move your new tests to the end? This really confused git's diff.

api/v1alpha1/package_webhook.go
```diff
@@ -0,0 +1,196 @@
...
+func InitPackageValidator(mgr ctrl.Manager) error {
+	mgr.GetWebhookServer().
+		Register("/validate-packages-eks-amazonaws-com-v1alpha1-package",
+			&webhook.Admission{Handler: &packageValidator{
+				Client: mgr.GetClient(),
+			}})
+	return nil
+}
+
+func (v *packageValidator) Handle(ctx context.Context, request admission.Request) admission.Response {
```
lewisdiamond, line 51: Is there a reason for not using kubebuilder's webhook annotation?

api/v1alpha1/webhook_suite_test.go
a-cool-train: Moved my tests to the end.

api/v1alpha1/package_webhook.go
a-cool-train: As discussed, I need access to resources from the cluster which is not possible to achieve with kubebuilder's webhook annotation.

TerryHowe (approved):
/lgtm

PR #314, opened 2022-07-21
Pbc upgrade available

TerryHowe:
Test coverage for bundle package is 98%

* Add upgradeAvailable state to PBC
* created kubeVersionClient and registryClient to make testing bundle state machines easier
* moved LatestBundle call out of controller to bundle state machine 
* moved PBC state machine to bundle package

Closes: https://github.com/aws/eks-anywhere-packages/issues/331


controllers/package_controller.go
```diff
@@ -80,9 +80,11 @@ func RegisterPackageReconciler(mgr ctrl.Manager) (err error) {
 	if err != nil {
 		return fmt.Errorf("creating discovery client: %s", err)
 	}
+	kvc := bundle.NewKubeVersionClient(discovery)
 	puller := artifacts.NewRegistryPuller()
+	registryClient := bundle.NewRegistryClient(log, puller)
 	bundleClient := bundle.NewPackageBundleClient(mgr.GetClient())
-	bundleManager := bundle.NewBundleManager(log, discovery, puller, bundleClient)
+	bundleManager := bundle.NewBundleManager(log, kvc, registryClient, bundleClient)
```
a-cool-train, line 87: Should we create a factory for this repo where all new creations happen in? The idea would be to do the instantiation of new objects in one place. We could follow a similar pattern from eks-anywhere. Similar to this: https://github.com/aws/eks-anywhere/blob/main/pkg/dependencies/factory.go. I found it very helpful for managing object creations.

controllers/packagebundle_controller_test.go
```diff
@@ -8,38 +8,58 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/golang/mock/gomock"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 
-	"github.com/aws/eks-anywhere-packages/api/v1alpha1"
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
```
a-cool-train, line 16: Might be better to call it something like packagesv1. Thoughts?

controllers/packagebundle_controller_test.go
```diff
@@ -8,38 +8,58 @@ import (
...
-			Namespace: v1alpha1.PackageNamespace,
+			Namespace: api.PackageNamespace,
 		},
 	}
 }
 
+func GivenBundle() *api.PackageBundle {
+	return &api.PackageBundle{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "v1-22-1001",
```
a-cool-train, line 34: Should we take the bundle name as an argument so this can be used for other tests?

controllers/packagebundle_controller_watch_test.go
```diff
@@ -29,7 +29,7 @@ func TestPackageBundleReconciler_mapBundleReconcileRequests(t *testing.T) {
 			return nil
 		})
 	bm := bundlefake.NewBundleManager()
-	sut := NewPackageBundleReconciler(mockClient, nil, mockBundleClient, bm, logr.Discard())
+	sut := NewPackageBundleReconciler(mockClient, nil, mockBundleClient, bm, nil, logr.Discard())
```
a-cool-train, line 32: Maybe let's try to use optional parameters similar to this: https://github.com/aws/eks-anywhere/blob/8bd0a318a0acd668c14166dfd9cac910653f347e/pkg/curatedpackages/packageclient.go#L29

controllers/packagebundlecontroller_controller_test.go
```diff
@@ -67,162 +66,130 @@ func TestPackageBundleControllerReconcilerReconcile(t *testing.T) {
 		}
 	}
 
-	t.Run("marks status inactive if name doesn't match", func(t *testing.T) {
+	t.Run("happy path", func(t *testing.T) {
```
a-cool-train, line 69: Can we be more descriptive for this test?

controllers/packagebundlecontroller_controller_test.go
```diff
@@ -67,162 +66,130 @@ func TestPackageBundleControllerReconcilerReconcile(t *testing.T) {
...
-		}
-		if result.Requeue {
-			t.Errorf("expected Requeue to be false, got true")
+			t.Errorf("expected no error, got %s", err)
 		}
+		assert.True(t, result.Requeue)
 	})
 
-	t.Run("leaves state as is", func(t *testing.T) {
+	t.Run("process error", func(t *testing.T) {
```
a-cool-train, line 90: Can we be more descriptive?

controllers/packagebundlecontroller_controller_test.go
```diff
@@ -67,162 +66,130 @@ func TestPackageBundleControllerReconcilerReconcile(t *testing.T) {
...
-		_, err := r.Reconcile(ctx, req)
+		result, err := r.Reconcile(ctx, req)
 		if err != nil {
 			t.Errorf("expected no error, got %s", err)
 		}
+		assert.True(t, result.Requeue)
 	})
 
-	t.Run("marks status disconnected if active", func(t *testing.T) {
+	t.Run("marks status ignored bogus name", func(t *testing.T) {
```
a-cool-train, line 111: This doesn't mention what was being tested and a little more description would be great.

controllers/packagebundlecontroller_controller_test.go
```diff
@@ -67,162 +66,130 @@ func TestPackageBundleControllerReconcilerReconcile(t *testing.T) {
...
+		result, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: ignoredController})
 		if err != nil {
-			t.Errorf("expected no error, got %s", err)
+			t.Fatalf("expected no error, got %s", err)
 		}
+		assert.False(t, result.Requeue)
 	})
 
-	t.Run("marks status active if disconnected", func(t *testing.T) {
+	t.Run("status error ignored bogus name", func(t *testing.T) {
```
a-cool-train, line 145: Be more descriptive on what this is supposed to achieve.

controllers/packagebundlecontroller_controller_test.go
```diff
@@ -67,162 +66,130 @@ func TestPackageBundleControllerReconcilerReconcile(t *testing.T) {
...
+		result, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: ignoredController})
 		if err != nil {
-			t.Errorf("expected no error, got %s", err)
+			t.Fatalf("expected no error, got %s", err)
 		}
+		assert.False(t, result.Requeue)
 	})
 
-	t.Run("marks status active if state not set", func(t *testing.T) {
+	t.Run("already ignored bogus name", func(t *testing.T) {
```
a-cool-train, line 171: Same as above comments on descriptiveness

generatebundlefile/ecr_helper.go
```diff
@@ -5,8 +5,9 @@ import (
 	"os/exec"
 	"strings"
 
-	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
 	"github.com/pkg/errors"
+
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
```
a-cool-train, line 10: Can we have a better name for the import?

pkg/bundle/client_test.go
```diff
@@ -44,16 +45,17 @@ func givenBundle() api.PackageBundle {
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      api.PackageBundleControllerName,
 			Namespace: api.PackageNamespace,
 		},
 		Spec: api.PackageBundleControllerSpec{
 			ActiveBundle: testBundleName,
 			Source: api.PackageBundleControllerSource{
-				Registry: "public.ecr.aws/j0a1m4z9",
+				Registry:   testBundleRegistry,
+				Repository: "eks-anywhere-package-bundles",
```
a-cool-train, line 58: Move this to a constant?

pkg/bundle/registry_client.go
```diff
@@ -0,0 +1,68 @@
...
+	LatestBundle(ctx context.Context, baseRef string, kubeVersion string) (
+		*api.PackageBundle, error)
+
+	// DownloadBundle downloads the bundle with a given tag.
+	DownloadBundle(ctx context.Context, ref string) (
+		*api.PackageBundle, error)
+}
+
+type registryClient struct {
+	log    logr.Logger
```
a-cool-train, line 27: As we discussed log is not required here. But if it is needed elsewhere please let's make it an optional parameter.

a-cool-train:
This will require some CLI changes.

Makefile
```diff
@@ -48,6 +48,7 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 .PHONY: fmt
 fmt: ## Run gofmt against code.
 	gofmt -s -w $(shell find . -name '*go' | grep -v mock)
+	gci write --skip-generated -s standard,default -s 'prefix(github.com/aws/eks-anywhere-packages)' $(shell find . -name '*go' | grep -v mock)
```
ewollesen, line 51: I promise I won't ask for too many things to be broken out into separate PRs, but this one I have to ask about. The reason being that this change is probably responsible for a bunch of small styling changes in this PR, and this PR is already large enough without me having to try and figure out if the changes are just styling.

Also, I wouldn't mind if this change were discussed as a quick post-standup discussion in case anyone has any feelings about the tools we're using. If possible, I'd love if we could pick a single one of the 18 billion go formatting tools out there, and just stick with it. :) I'm exaggerating a little, but honestly, I can't think of a single thing that one formatter has over another that actually matters to me. Others may feel differently of course.

ewollesen (commented):
Just a review of 1 comment for now, I'll keep going, but I wanted you to be able to see this one quickly.

Makefile
TerryHowe: That was to be compliant with github actions

controllers/packagebundlecontroller_controller_test.go
ewollesen: It's a common term in test scenarios. It means "everything goes as expected and according to plan". It's also sometimes called "golden path" or "smoke test". It's just a shorthand.

controllers/packagebundlecontroller_controller_test.go
```diff
@@ -67,162 +66,130 @@ func TestPackageBundleControllerReconcilerReconcile(t *testing.T) {
...
-				return nil
-			})
+		mockBundleManager := bundleMocks.NewMockManager(gomock.NewController(t))
+		mockBundleManager.EXPECT().ProcessBundleController(ctx, &pbc).Return(nil)
 
-		r := NewPackageBundleControllerReconciler(mockClient, nil, bm,
+		r := NewPackageBundleControllerReconciler(mockClient, nil, mockBundleManager,
 			logr.Discard())
-		result, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: inactiveController})
+		result, err := r.Reconcile(ctx, req)
```
ewollesen, line 83: If you want to use testify (and I'm for that), you could replace this with:
```
assert.NoError(t, err)
```
That way this function uses testify all the way through, instead of half-and-half. You're probably aware, but the functions in testify's `assert` package have the same semantics as `t.Errorf`, as in they continue the execution of the test. The functions in testify's `require` package correlate to `t.Fatalf`.

pkg/bundle/kube_version.go
```diff
@@ -0,0 +1,37 @@
+package bundle
+
+import (
+	_ "embed"
```
ewollesen, line 4: I'm a big fan of the embed package, but I see no use of it, or its directives. Are tehse necessary?

```sh
$ rg -u embed
pkg/bundle/registry_client.go
6:	_ "embed"

pkg/bundle/manager.go
5:	_ "embed"

pkg/bundle/kube_version.go
4:	_ "embed"

$ rg -u go:embed
<no results>
```

pkg/bundle/kube_version.go
```diff
@@ -0,0 +1,37 @@
...
+type KubeVersionClient interface {
+	// ApiVersion returns the Kubernetes API version
+	ApiVersion() (string, error)
+}
+
+type kubeVersionClient struct {
+	kubeServerVersion discovery.ServerVersionInterface
+}
+
+func NewKubeVersionClient(serverVersion discovery.ServerVersionInterface) (manager *kubeVersionClient) {
```
ewollesen, line 20:
```suggestion
func NewKubeVersionClient(serverVersion discovery.ServerVersionInterface) (*kubeVersionClient) {
```

Nitpick really, but naming return values is best avoided, unless you're documenting what they are via names. The names get stale, or allow naked returns, which are best avoided.

pkg/bundle/kube_version.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	_ "embed"
+	"fmt"
+	"strings"
+
+	"k8s.io/client-go/discovery"
+)
+
+type KubeVersionClient interface {
```
ewollesen, line 11: I wonder if this interface is very useful? It's only got the smallest of changes compared to discovery.ServerVersionInterface, I just don't think it's worth an additional struct, interface, and all that goes with it.

The formatting changes you want for the version can easily be implemented as a pure function: 
```go
formatKubernetesVersion(major, minor string) string {}
```
Adding this interface brings no testability gains, as you can just as easily implement a mock of discovery.ServerVersionInterface as you can one of KubeVersionClient. (I could swear I had written one of those at some point...)

ewollesen (commented):
I'm posting these now, lest they get lost overnight, I'll pick up my review in the morning. Please don't worry about these comments much yet though, I may edit them tomorrow, as I still consider them all drafts.

generatebundlefile/ecr_helper.go
TerryHowe: This was fixed in another pr

pkg/bundle/manager.go
```diff
@@ -144,84 +127,88 @@ func (m bundleManager) SortBundlesDescending(bundles []api.PackageBundle) {
...
+		m.log.V(6).Info("update", "PackageBundleController", pbc.Name, "state", pbc.Status.State)
+		err = m.bundleClient.SaveStatus(ctx, pbc)
+		if err != nil {
+			return fmt.Errorf("updating %s status to %s: %s", pbc.Name, pbc.Status.State, err)
+		}
+	case "":
+		if pbc.Spec.ActiveBundle == "" {
+			pbc.Spec.ActiveBundle = latestBundle.Name
+		}
+		pbc.Status.State = api.BundleControllerStateActive
```
ewollesen, line 205: Does this change get applied?

> The PUT and POST verbs on objects MUST ignore the status values, to avoid accidentally overwriting the status in read-modify-write scenarios. A /status subresource MUST be provided to enable system components to update statuses of resources they manage.
> -- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

It's possible this isn't having an effect? Though I guess, it'll get picked up at the next reconcile cycle anyway? But this might need it's own specific SaveStatus call?

pkg/bundle/manager.go
```diff
@@ -144,84 +127,88 @@ func (m bundleManager) SortBundlesDescending(bundles []api.PackageBundle) {
...
+			return fmt.Errorf("updating %s status to %s: %s", pbc.Name, pbc.Status.State, err)
+		}
+	case api.BundleControllerStateDisconnected:
+		pbc.Status.State = api.BundleControllerStateActive
+		m.log.V(6).Info("update", "PackageBundleController", pbc.Name, "state", pbc.Status.State)
+		err = m.bundleClient.SaveStatus(ctx, pbc)
+		if err != nil {
+			return fmt.Errorf("updating %s status to %s: %s", pbc.Name, pbc.Status.State, err)
+		}
+	case "":
```
ewollesen, line 201:
```suggestion
	default:
```

ewollesen (changes requested):

ewollesen:
@TerryHowe you should also check how these changes effect eks-anywhere (specifically the CLI), since I suspect there will be some breakage for them. You might go back over these changes with an eye to exactly what is exported too, lots of function in here seem to be exported for no particularly good reason, those are all liabilities that we have to check whenever we ~break things~ change their behavior. :)

Makefile
TerryHowe: I'm throwing this change to https://github.com/aws/eks-anywhere-packages/pull/326 to investigate alternatives later

Makefile
```diff
@@ -158,19 +159,27 @@ endef
...
+		$(MOCKGEN) -source pkg/bundle/manager.go -destination=pkg/bundle/mocks/manager.go -package=mocks Manager
+
+pkg/bundle/mocks/registry_client.go: pkg/bundle/registry_client.go
+	PATH=$(shell $(GO) env GOROOT)/bin:$$PATH \
+		$(MOCKGEN) -source pkg/bundle/registry_client.go -destination=pkg/bundle/mocks/registry_client.go -package=mocks RegistryClient
 
 pkg/packages/mocks/manager.go: pkg/packages/manager.go
 	PATH=$(shell $(GO) env GOROOT)/bin:$$PATH \
-		$(MOCKGEN) -source pkg/packages/manager.go -destination=pkg/packages/mocks/manager.go -package=mocks Manager
+		$(MOCKGEN) -source pkg/packages/manager.go -destination=pkg/packages/mocks/registrmanagery_client.go -package=mocks Manager
```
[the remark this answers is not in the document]
TerryHowe: Yeh, how did that even happen I wonder. Fixed.

controllers/packagebundlecontroller_controller.go
```diff
@@ -109,53 +111,20 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 			r.Log.V(6).Info("update", "PackageBundleController", pbc.Name, "state", pbc.Status.State)
 			err = r.Client.Status().Update(ctx, pbc, &client.UpdateOptions{})
 			if err != nil {
-				return withoutRequeue(result), fmt.Errorf("updating status to ignored: %s", err)
+				r.Log.Error(err, "updating ignored status")
```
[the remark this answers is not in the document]
TerryHowe: I like to only bubble up errors that are either the fault of the caller or they can do something about.  All that happens when we return an error here is they log it and the message logged is harder to read than this one.

controllers/packagebundlecontroller_controller_test.go
TerryHowe: I strongly agree. In another PR though.

controllers/package_controller.go
TerryHowe: IMO, the CLI took dependency injection a bit too far and that is more of the symptom of the problem rather than a solution. Factories are supposed to hide the implementation and that really doesn't do that. I think creating that factory is a good step for them in resolving the issue, but it is just a step.

pkg/bundle/manager.go
```diff
@@ -144,84 +127,88 @@ func (m bundleManager) SortBundlesDescending(bundles []api.PackageBundle) {
...
-		return nil, fmt.Errorf("package bundle artifact is empty")
-	}
-
-	bundle := &api.PackageBundle{}
-	err = yaml.Unmarshal(data, bundle)
-	if err != nil {
-		return nil, fmt.Errorf("unmarshalling package bundle: %s", err)
+		m.log.Error(err, "Unable to get latest bundle")
+		if pbc.Status.State == api.BundleControllerStateActive {
+			m.log.Error(err, "marking disconnected")
```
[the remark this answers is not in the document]
TerryHowe: I looked into it and the error actually isn't logged until this point.

pkg/bundle/manager.go
```diff
@@ -144,84 +127,88 @@ func (m bundleManager) SortBundlesDescending(bundles []api.PackageBundle) {
...
-
-	if len(bytes.TrimSpace(data)) == 0 {
-		return nil, fmt.Errorf("package bundle artifact is empty")
-	}
-
-	bundle := &api.PackageBundle{}
-	err = yaml.Unmarshal(data, bundle)
-	if err != nil {
-		return nil, fmt.Errorf("unmarshalling package bundle: %s", err)
+		m.log.Error(err, "Unable to get latest bundle")
```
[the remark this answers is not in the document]
TerryHowe: Yeh, this is a little long, but I think it looks a lot worse in the diffs. I'd like to leave this idea for a future PR.

pkg/bundle/kube_version.go
TerryHowe: Good question, but I'd like to leave this as is for now as this is a bit easier thing for the CLI to use.

pkg/bundle/manager.go
TerryHowe: Done. Three reconcile loops caused by this, I don't think it could be done in less.
```
 % k get -n eksa-packages packages,packagebundles,packagebundlecontroller
NAME                                                STATE
packagebundle.packages.eks.amazonaws.com/v1-21-33   active

NAME                                                                                 ACTIVEBUNDLE   STATE    DETAIL
packagebundlecontroller.packages.eks.amazonaws.com/eksa-packages-bundle-controller   v1-21-33       active
```

controllers/packagebundlecontroller_controller.go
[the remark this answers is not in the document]
ewollesen: I see your point about the returned error, when logged, being harder to read.

I still would've chosen something other than Error here, as I think that's overstating things, but I suppose it's not a big deal either way.

pkg/bundle/kube_version.go
ewollesen: The CLI isn't using it right now is it? Given that it's new, I suppose it can't be using it at the moment... So why introduce it at all for the CLI?

pkg/bundle/manager.go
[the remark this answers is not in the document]
ewollesen: It was logged on line 137 (2 lines above this one). "Unable to get latest bundle". Or am I missing something?

controllers/packagebundlecontroller_controller.go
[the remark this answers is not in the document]
TerryHowe: the log error dumps the stack trace as I recall. I think that is why I was using it.

pkg/bundle/kube_version.go
TerryHowe: done

pkg/bundle/manager.go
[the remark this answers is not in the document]
TerryHowe:
```suggestion
```
Not sure what I was looking at

pkg/bundle/kube_version.go
ewollesen: https://github.com/TerryHowe/eks-anywhere-packages/pull/1

Makefile
```diff
@@ -158,15 +158,23 @@ endef
...
+mocks: mockgen controllers/mocks/client.go controllers/mocks/manager.go pkg/driver/mocks/packagedriver.go pkg/bundle/mocks/client.go pkg/bundle/mocks/manager.go pkg/bundle/mocks/registry_client.go pkg/packages/mocks/manager.go
 
 pkg/bundle/mocks/client.go: pkg/bundle/client.go
 	PATH=$(shell $(GO) env GOROOT)/bin:$$PATH \
 		$(MOCKGEN) -source pkg/bundle/client.go -destination=pkg/bundle/mocks/client.go -package=mocks Client
 
+pkg/bundle/mocks/kube_version.go: pkg/bundle/kube_version.go
+	PATH=$(shell $(GO) env GOROOT)/bin:$$PATH \
+		$(MOCKGEN) -source pkg/bundle/kube_version.go -destination=pkg/bundle/mocks/kube_version.go -package=mocks KubeVersion
+
```
ewollesen, line 170:
```suggestion
```
I missed this in my PR. :(

ewollesen (approved):
I found one place where a few lines should have been removed in my PR, then I think it'll be g2g.

PR #315, opened 2022-07-25
Feature add schema to bundle

a-cool-train:
*Description of changes:*
Add json schema to bundle.


api/v1alpha1/packagebundle_types.go
```diff
@@ -94,19 +94,6 @@ type VersionImages struct {
 	Digest string `json:"digest"`
 }
```
TerryHowe, line 96: Do we need a period of backward compatibility for this and just leve the version config until everything in build tooling uses schema?

a-cool-train: I think there will be a breaking change at some point, I'd rather it be before GA than after GA.

TerryHowe: I was thinking if you left the config parsing, it would not fail parsing that contains a config

TerryHowe (approved):
/lgtm

PR #318, opened 2022-07-26
ECR Token Refresh Application

junshun:
*Issue #317

*Description of changes:*
Go Program that supports reading AWS credentials via environment variables or Assuming a web identity (for uses cases such as IAM for service Accounts). This binary is intended to built into a container and run as a cronjob to keep ECR Tokens up to date.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #317, opened 2022-07-26
Private ECR Registry support


junshun:
Need support for pulling images from private ECR registry and address issue of ECR token timeout every 12 hours.

TerryHowe:
Done

PR #324, opened 2022-07-28
Updating helmchart to add template for ecr-refresh cronjob

junshun:
*Issue #317

*Description of changes:*
Adding in helm chart for deploying cronjob. This is meant to work with a version #318 that is running in a container.

### DIFF
--- a/charts/eks-anywhere-packages/templates/cronjob.yaml
+++ b/charts/eks-anywhere-packages/templates/cronjob.yaml
@@ -4,7 +4,7 @@ metadata:
   namespace: {{ .Values.namespace }}
   name: cron-ecr-renew
 spec:
-  schedule: "0 */6 * * *"
+  schedule: "0 */5 * * *"
   successfulJobsHistoryLimit: 3
   failedJobsHistoryLimit: 5
   jobTemplate:
@@ -15,15 +15,10 @@ spec:
           serviceAccountName: {{ include "eks-anywhere-packages.serviceAccountName" . }}
           containers:
             - name: {{.Values.cronjob.name}}
-              image: {{.Values.sourceRegistry}}{{ template "template.image" .Values.cronjob }}
+              image: {{.Values.sourceRegistry}}{{.Values.cronjob.tag}}
               imagePullPolicy: {{ .Values.imagePullPolicy }}
               env:
-                - name: TARGET_NAMESPACE
-                  valueFrom:
-                    configMapKeyRef:
-                      name: {{.Values.nsconfigmap.name}}
-                      key: NAMESPACES
-                - name: DOCKER_SECRET_NAME
+                - name: ECR_TOKEN_SECRET_NAME
                   value: ecr-token
                 - name: AWS_ACCESS_KEY_ID
                   valueFrom:
--- a/charts/eks-anywhere-packages/templates/cronjob.yaml
+++ b/charts/eks-anywhere-packages/templates/cronjob.yaml
@@ -15,7 +15,7 @@ spec:
           serviceAccountName: {{ include "eks-anywhere-packages.serviceAccountName" . }}
           containers:
             - name: {{.Values.cronjob.name}}
-              image: {{.Values.sourceRegistry}}{{.Values.cronjob.tag}}
+              image: {{.Values.sourceRegistry}}{{ template "template.image" .Values.cronjob }}
               imagePullPolicy: {{ .Values.imagePullPolicy }}
               env:
                 - name: ECR_TOKEN_SECRET_NAME
--- a/charts/eks-anywhere-packages/templates/cronjob.yaml
+++ b/charts/eks-anywhere-packages/templates/cronjob.yaml
@@ -1,0 +1,45 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  namespace: {{ .Values.namespace }}
+  name: cron-ecr-renew
+spec:
+  schedule: "0 */6 * * *"
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 5
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: OnFailure
+          serviceAccountName: {{ include "eks-anywhere-packages.serviceAccountName" . }}
+          containers:
+            - name: {{.Values.cronjob.name}}
+              image: {{.Values.sourceRegistry}}{{ template "template.image" .Values.cronjob }}
+              imagePullPolicy: {{ .Values.imagePullPolicy }}
+              env:
+                - name: TARGET_NAMESPACE
+                  valueFrom:
+                    configMapKeyRef:
+                      name: {{.Values.nsconfigmap.name}}
+                      key: NAMESPACES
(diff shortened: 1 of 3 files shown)


charts/eks-anywhere-packages/values.yaml
```diff
@@ -85,4 +85,14 @@ controller:
       memory: 50Mi
     limits:
       cpu: 500m
-      memory: 300Mi
\ No newline at end of file
+      memory: 300Mi
+# Additional Variables to configure cronjob
+cronjob:
+  name: ecr-refresher
+  # -- ECR refresher repository name.
```
TerryHowe, line 92: Won't we be targeting this to another repository?

TerryHowe (changes requested):
We can't really merge this until we have a build at least in dev

TerryHowe:
As soon as the container image is being built, we can add these changes with the helm chart or we'd break the helm chart

charts/eks-anywhere-packages/templates/cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+            - name: {{.Values.cronjob.name}}
+              image: {{.Values.sourceRegistry}}{{ template "template.image" .Values.cronjob }}
+              imagePullPolicy: {{ .Values.imagePullPolicy }}
+              env:
+                - name: TARGET_NAMESPACE
+                  valueFrom:
+                    configMapKeyRef:
+                      name: {{.Values.nsconfigmap.name}}
+                      key: NAMESPACES
+                - name: DOCKER_SECRET_NAME
```
TerryHowe, line 26: I don't think DOCKER in the name is great, but I'm not entirely sure how this is used.
```suggestion
                - name:ECR_TOKEN_SECRET_NAME
```

charts/eks-anywhere-packages/templates/cronjob.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  namespace: {{ .Values.namespace }}
```
TerryHowe, line 4: I almost wonder if these secrets and this cron job should run in the eksa-system namespace instead of eksa-packages

junshun: The reasoning for running this in eksa-packages is the configmap that keeps track of namespaces that the controller installs to.

charts/eks-anywhere-packages/templates/cronjob.yaml
junshun: I am fine with this change reason I call it docker, as the secret type is kubernetes.io/dockerconfigjson even though its not really a docker specific type.

charts/eks-anywhere-packages/templates/cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+spec:
+  schedule: "0 */6 * * *"
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 5
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: OnFailure
+          serviceAccountName: {{ include "eks-anywhere-packages.serviceAccountName" . }}
```
jonahjon, line 15: Should this have a different service account with different K8s rbac since our actual package controller doesn't have the same requirements to write configmaps/secrets? I would be ok with this for now, since this is temporary even though it's the best.

charts/eks-anywhere-packages/values.yaml
```diff
@@ -85,4 +85,14 @@ controller:
...
+  repository: "eks-anywhere-packages"
```
TerryHowe, line 93:
```suggestion
  repository: "packages-token-refresher"
```

junshun: Do we need build tooling to also push this to a seperate repository?

TerryHowe: yes and cdk to create the repo, Jonah to make sure it gets promoted.

ivyostosh: This should be:
`repository: "eks-anywhere-packages/ecrtokenrefresher"` per [build tooling PR](https://github.com/aws/eks-anywhere-build-tooling/pull/1107). I can re-name it to something else, but I was hoping to use this structure to be clear that this image is built from the `eks-anywhere-packages` project.

ivyostosh: Ok. Then the update here will be:
`repository: "eks-anywhere-packages/ecr-token-refresher"`
I will update the build tooling PR.

TerryHowe: Seems really odd to me to create a repository that has the path of another repository in it. I imagine it will work though.

charts/eks-anywhere-packages/templates/cronjob.yaml
junshun: I believe the package controller already has ability to write configmaps. As for secrets the package controller here only consumes secrets. Helm is actually the one that is creating the secret in other namespaces via passing in the value to the helm charts

charts/eks-anywhere-packages/values.yaml
junshun: Updated to different repository

charts/eks-anywhere-packages/values.yaml
junshun: Changed name to match

charts/eks-anywhere-packages/values.yaml
```diff
@@ -85,4 +85,14 @@ controller:
...
     limits:
       cpu: 500m
-      memory: 300Mi
\ No newline at end of file
+      memory: 300Mi
+# Additional Variables to configure cronjob
+cronjob:
+  name: ecr-refresher
+  # -- ECR refresher repository name.
+  repository: "eks-anywhere-packages/ecr-token-refresher"
```
TerryHowe, line 93: seriously, this format might not be compatible with all registries where we have a repository named `eks-anywhere-packages` and another named `eks-anywhere-packages/ecr-token-refresher` logically it makes no sense

charts/eks-anywhere-packages/values.yaml
ivyostosh: I was trying to follow the pattern of `org/project`, which was used by many projects in the build-tooling. But based on today's conversation, I don't think we want to move forward with this naming convention anyway (though we still don't have a conclusion on this yet). Given this is a short-term solution, I'm thinking we can update it to something like `repository: "ecr-token-refresher"` or something similar (whatever makes sense logically).

Just let me know what you guys decide on, and I will update build-tooling accordingly.

charts/eks-anywhere-packages/values.yaml
ivyostosh: Responded in an earlier comment. No strong preference on my end.

junshun: going with ecr-token-refresher .

TerryHowe (approved):
/lgtm

PR #326, opened 2022-07-29
Add gci write to make fmt

TerryHowe:
I think we may need this, but need to investigate alternatives first. This does work.

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,10 +53,6 @@ SRC_FILES = $(shell git ls-files --exclude-standard | grep '\.go$' | grep -v '/m
 .PHONY: fmt
 fmt: run-gofmt run-gci
 
-.PHONY: ls-go-files
-ls-go-files:
-	@
-
 LS_FILES_CMD = git ls-files --exclude-standard | grep '\.go$$' | grep -v '/mocks/\|zz_generated\.'
 
 .PHONY: run-gofmt
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,9 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: fmt
 fmt: ## Run gofmt against code.
-	gofmt -s -w $(shell find . -name '*go' | grep -v mock)
+	$(GO) install github.com/daixiang0/gci@v0.6.3
+	find . -name '*go' | grep -v mock | xargs gofmt -s -w
+	find . -name '*go' | grep -v mock | xargs gci write --skip-generated -s standard,default -s 'prefix(github.com/aws/eks-anywhere-packages)'
 
 .PHONY: lint
 lint: bin/golangci-lint ## Run golangci-lint
--- a/Makefile
+++ b/Makefile
@@ -48,8 +48,6 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 $(GOBIN)/gci:
 	$(GO) install github.com/daixiang0/gci@v0.6.3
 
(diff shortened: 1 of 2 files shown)


ewollesen:
As I recall, you said you did this to make a github action happy?

The github action uses [golangci-lint](https://github.com/golangci/golangci-lint).

It seems like asking for trouble to bet that gci and golangci-lint will have the same output in all cases, so if your goal is trying to satisfy the github action, then I'd suggest matching the version of golangci-lint that the github action uses.

That said, I don't care much which one we use, but I would greatly prefer to just use one of them. There are a number of formatters and linters out there, but I can't really tell any difference in quality between them. As a result, I'd prefer to simply pick one and stick to it, rather than adding more on top of more. In my experience that just causes more problems with versions needing me line up, and things being run in the correct order, and it's harder to run it from your editor, etc.

https://github.com/golangci/golangci-lint
https://pkg.go.dev/cmd/gofmt
https://github.com/daixiang0/gci
https://pkg.go.dev/golang.org/x/tools/cmd/goimports
https://staticcheck.io/
https://revive.run/
https://github.com/golang/lint

WOW. There are entire lists of linters now: https://github.com/golangci/awesome-go-linters#linters

TerryHowe:
> 


As far as I know goangci-lint does not format. You have a million ideas, but no solution.

ewollesen:
> As far as I know goangci-lint does not format. You have a million ideas, but no solution.

I apologize, I was under the impression that golangci-lint performed formatting as well as static analysis. I was wrong.

While running gci isn't going to fix all of the possible things golangci-lint will complain about, it could certainly help fix some of them for us.

That said, I'm unable to run `make fmt` anyway:
```
make: execvp: /usr/bin/env: Argument list too long
```

Perhaps it's a Mac/Linux thing?

I suspect there might also be issues with the `gci` line, which has several flags that don't appear in my `gci --help` output. Perhaps a note could be added that describes the version required, or installs it? (Though gci --version doesn't seem to be a valid flag, so I don't really know which version of gci I have)

Apologies again for the aggravation caused.

Makefile
```diff
@@ -45,9 +45,27 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
...
+SRC_FILES = $(shell git ls-files --exclude-standard | grep '\.go$' | grep -v '/mocks/\|zz_generated\.')
+
 .PHONY: fmt
-fmt: ## Run gofmt against code.
-	gofmt -s -w $(shell find . -name '*go' | grep -v mock)
+fmt: run-gofmt run-gci
+
+.PHONY: ls-go-files
+ls-go-files:
+	@
```
ewollesen, line 58:
```suggestion
```
Oops, I should review my own code a little more closely.

Makefile
```diff
@@ -45,9 +45,23 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+$(GOBIN)/gci:
+	$(GO) install github.com/daixiang0/gci@v0.6.3
+
+SRC_FILES = $(shell git ls-files --exclude-standard | grep '\.go$' | grep -v '/mocks/\|zz_generated\.')
+
```
TerryHowe, line 52:
```suggestion
```

ewollesen (commented):
/lgtm

PR #335, opened 2022-08-02
Adding ability for HelmDriver to pass in ImagePullSecrets to charts

junshun:
*Issue #317

*Description of changes:*
Adding the ability for helmdriver to pass in ecr imagePullSecrets for helm charts. 
Added functionaitly to get ecrsecret and update config map with installed namespaces to be used by cronjob for refreshing tokens.


pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type ecrSecret struct {
+	clientset kubernetes.Interface
+}
+
+var _ Authenticator = (*ecrSecret)(nil)
+
+func NewECRSecret(config *rest.Config) *ecrSecret {
+	clientset, _ := kubernetes.NewForConfig(config)
```
jonahjon, line 29: What happens if this fails?

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (s *ecrSecret) AuthFilename() (string, error) {
+	// Check if Helm registry config is set
+	helmconfig := os.Getenv("HELM_REGISTRY_CONFIG")
+	if helmconfig != "" {
+		// Use HELM_REGISTRY_CONFIG
+		return helmconfig, nil
+	}
+
+	return "", nil
```
jonahjon, line 44: Shouldn't we return an error not nil? Otherwise remove the error from the return types.

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,110 @@
...
+		Update(ctx, cm, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func addToConfigMap(cm *v1.ConfigMap, namespace string, add bool) {
+	if namespace == "" {
+		namespace = "default"
```
jonahjon, line 64: So if a namespace isn't specified we "default" is what we add to the configmap not `eksa-packages` just trying to clarify

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return nil, err
+	}
+	values := make(map[string]interface{})
+	var imagePullSecret [1]map[string]string
+	imagePullSecret[0] = make(map[string]string)
+	imagePullSecret[0]["name"] = varECRTokenName
+	values["imagePullSecrets"] = imagePullSecret
+	values["pullSecretName"] = varECRTokenName
+	values["pullSecretData"] = b64.StdEncoding.EncodeToString(secret.Data[".dockerconfigjson"])
```
jonahjon, line 108: It feels like we should have some soft of check whether this file exists and return an error here, unless the b64.StdEncoding will naturally error if the file doesn't exist.

pkg/driver/helmdriver.go
```diff
@@ -70,12 +73,25 @@ func (d *helmDriver) Install(ctx context.Context,
 		return fmt.Errorf("loading helm chart %s: %w", name, err)
 	}
 
+	// Update values with imagePullSecrets
+	secretvals, err := d.secretAuth.GetSecretValues(ctx)
+	if err == nil {
```
jonahjon, line 78: What if err != nil?

jonahjon (changes requested):

pkg/driver/helmdriver.go
junshun: So in cases where their are enviroments that don't need imagePullSecrets, such as registry mirroring to a local registry/or packages in a public registry we want to helm install to continue but not populate the imagePullSecrets value in the chart.

pkg/authenticator/ecrsecret.go
junshun: Good Catch, added check eariler as if secret doesn't exist we don't need to fill out the other value fields either.

pkg/authenticator/ecrsecret.go
junshun: Yes, though this is changed now to expect a defaultNamespace and use "default" if that is not set

pkg/authenticator/ecrsecret.go
junshun: added error check

pkg/authenticator/ecrsecret.go
junshun: Removed

pkg/authenticator/ecrsecret_test.go
```diff
@@ -0,0 +1,222 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/rest"
+)
+
+func TestAuthFilename(t *testing.T) {
+	config := rest.Config{}
+	t.Run("golden path for set HELM_REGISTRY_CONFIG", func(t *testing.T) {
+		testfile := "/test.txt"
+		os.Setenv("HELM_REGISTRY_CONFIG", testfile)
```
ewollesen, line 20: Please use the [testing version](https://cs.opensource.google/go/go/+/go1.19:src/testing/testing.go;l=1287), so it's cleaned up afterwards.

pkg/authenticator/ecrsecret_test.go
```diff
@@ -0,0 +1,222 @@
...
+		os.Setenv("HELM_REGISTRY_CONFIG", "")
+		ecrAuth, _ := NewECRSecret(&config)
+		val := ecrAuth.AuthFilename()
+
+		assert.Equal(t, val, "")
+	})
+}
+
+func TestUpdateConfigMap(t *testing.T) {
+	ctx := context.TODO()
```
ewollesen, line 37:
```suggestion
	ctx := context.Background()
```

pkg/authenticator/ecrsecret_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Name:      varConfigMapName,
+				Namespace: varPackagesNamespace,
+			},
+			Data: cmdata,
+		})
+		ecrAuth := ecrSecret{clientset: mockClientset}
+
+		err := ecrAuth.UpdateConfigMap(ctx, name, namespace, add)
+
+		assert.Nil(t, err)
```
ewollesen, line 56:
```suggestion
		err := ecrAuth.UpdateConfigMap(ctx, name, namespace, add)
		require.NoError(t, err)
```
NoError because it's clearer what you expect. "require", because it will stop the test if the condition doesn't hold. If you have an error there, the next call will almost certainly fail as well, so why litter the test output with more than necessary.

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+
+var _ Authenticator = (*ecrSecret)(nil)
+
+func NewECRSecret(config *rest.Config) (*ecrSecret, error) {
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	cm, err := clientset.CoreV1().ConfigMaps(varPackagesNamespace).
+		Get(context.TODO(), varConfigMapName, metav1.GetOptions{})
```
ewollesen, line 35: This constructor is doing a lot of things.

In general, the purpose of a constructor is to initialize any required memory fields within its struct. It's not really meant to be making network calls, or performing advanced logic. That sort of thing is best put in some sort of Init() method or the like.

The thing returned by a constructor should be safe from nil-pointer dereferences, not necessarily "ready to use", if that makes sense.

In this case, I suspect it might be better to let the caller of this initializer perform this ConfigMaps call, and pass the results into the constructor.

If it's too complicated or nuanced a thing for the caller to do, consider having another function, perhaps "SimpleSecretsAuthenticator" or something that could combine a few calls as necessary.

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+	// Store data
+	s.nsReleaseMap = cm.Data
+
+	return nil
+}
+
+func addToConfigMap(cm *v1.ConfigMap, name string, namespace string, add bool) {
```
ewollesen, line 74: The `add` parameter here is misleading. If it's set to false, then what happens is the value is removed. That conflicts with the name of the function.

Break this into two functions, named for what they do. 

If you really have to have one function, call it `modifyConfigMap` and pass an "operation" enum value of either ADD or REMOVE or something.

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	values := strings.Split(cm.Data[namespace], ",")
+	if add {
+		values = append(values, name)
+	}
+
+	result := make([]string, 0, len(values))
+	check := map[string]bool{}
+	for _, val := range values {
+		_, ok := check[val]
+		if !ok {
```
ewollesen, line 84:
```suggestion
		_, exists := check[val]
		if !exists {
```
`ok` is fine, when you're going to handle an error when `!ok`. But here it's just confusing.

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	update := ""
+	for _, val := range result {
+		if len(update) == 0 {
+			update = val
+		} else {
+			update = update + "," + val
+		}
+	}
```
ewollesen, line 101:
```suggestion
update := strings.Join(result, ",")
```
See https://pkg.go.dev/strings for lots of handy functions.

ewollesen (changes requested):

pkg/driver/helmdriver.go
```diff
@@ -69,13 +72,37 @@ func (d *helmDriver) Install(ctx context.Context,
...
+	// If no target namespace provided read chart values to find namespace
+	if namespace == "" {
+		val, exists := helmChart.Values["defaultNamespace"]
+		if exists {
+			namespace = val.(string)
+		} else {
+			// Fall back case of assuming its default
+			namespace = "default"
+		}
+	}
```
ewollesen, line 84:
```suggestion
	namespace := "default"
	if chartNS, ok := helmChart.Values["defaultNamespace"]; ok {	   
	   namespace = chartNS.(string)
	}
```
Shorter, and more idiomatic.

pkg/driver/helmdriver.go
```diff
@@ -69,13 +72,37 @@ func (d *helmDriver) Install(ctx context.Context,
...
+
+	// Update values with imagePullSecrets
+	// If no secret values we should still continue as it could be case of public registry or local registry
+	secretvals, err := d.secretAuth.GetSecretValues(ctx, namespace)
+	if err == nil {
+		for key, val := range secretvals {
+			values[key] = val
+		}
+	}
```
ewollesen, line 93: This is highly unusual for go code.

So much so, that I'd rather see this:

```suggestion
	secretvals, err := d.secretAuth.GetSecretValues(ctx, namespace)
	if err != nil {	
		// possibly a log.Warnf("something smoething")
		secretvals = nil
		// Continue, even though we failed to load secrets (or some words that makes sense to you)
	}

	// (remove me, this is for edification) The nil value will just skip right through...
	for key, val := range secretvals {
		values[key] = val
	}

```
Most people are just trained to see a call, then an error check with `err != nil`. If you mix it up and call `err == nil` on them, you should have a very good reason, and probably a comment that attempts to alert them. In this case, we do have an error, just a non-fatal one, and we can log that or otherwise deal with it, while keeping to the more common pattern of `err != nil`.

pkg/driver/helmdriver.go
```diff
@@ -69,13 +72,37 @@ func (d *helmDriver) Install(ctx context.Context,
...
 	_, err = get.Run(name)
 	if err != nil {
 		if errors.Is(err, driver.ErrReleaseNotFound) {
-			return d.createRelease(ctx, install, helmChart, values)
+			err = d.createRelease(ctx, install, helmChart, values)
+			if err == nil {
+				// Update installed-namespaces on successful install
+				err = d.secretAuth.UpdateConfigMap(ctx, name, namespace, true)
+			}
+			return err
```
ewollesen, line 105:
```suggestion
			err = d.createRelease(ctx, install, helmChart, values)
			if err != nil {
				return err
			}
			if err := d.secretAuth.UpdateConfigMap(ctx, name, namespace, true); err != nil {
				log.Warnf("failed to update namespace mapping or other words that make sense")
			}
			return nil // yeah, nil, the operation succeeded, just the bookkeeping failed. If this is important enough to fail the operation, then it should probalby have an inversion of control
```

pkg/driver/helmdriver.go
```diff
@@ -15,26 +15,28 @@ import (
...
 var _ PackageDriver = (*helmDriver)(nil)
 
-func NewHelm(log logr.Logger) (*helmDriver, error) {
+func NewHelm(log logr.Logger, k8sconfig *rest.Config) (*helmDriver, error) {
 	settings := cli.New()
 
 	// TODO Catch error here if not provided docker config or continue without an authfile if possible
-	secretAuth := auth.NewHelmSecret()
-	authfile, _ := secretAuth.AuthFilename()
+	secretAuth, _ := auth.NewECRSecret(k8sconfig)
```
ewollesen, line 38: I'd prefer to keep this kind of initialization out of the helm driver's constructor. Also, please always check your errors, don't leave them as a TODO.

pkg/driver/helmdriver.go
```diff
@@ -15,26 +15,28 @@ import (
...
+	authfile := secretAuth.AuthFilename()
```
ewollesen, line 39: I see no reason for this to be done here. What concern is this of the helm driver's?

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &ecrSecret{
+		clientset:    clientset,
+		nsReleaseMap: cm.Data,
+	}, nil
+}
+
+func (s *ecrSecret) AuthFilename() string {
```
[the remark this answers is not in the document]
junshun: So I wanted to keep this function here only because it is possible if there is a move to private helm chart this library can then use the ecr-token secret to generate an authfile that can then be used by helm driver. This would be an implementation that is unique to this and not in the helm driver library.

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	varPackagesNamespace = "eksa-packages"
```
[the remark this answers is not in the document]
junshun: done

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	varConfigMapName     = "ns-secret-map"
+	varECRTokenName      = "ecr-token"
```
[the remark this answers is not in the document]
junshun: done

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	values := make(map[string]interface{})
+	var imagePullSecret [1]map[string]string
+	imagePullSecret[0] = make(map[string]string)
+	imagePullSecret[0]["name"] = varECRTokenName
+	values["imagePullSecrets"] = imagePullSecret
+
+	// if namespace doesn't already have the secret we will fill out the secret.ymal
+	if _, exist := s.nsReleaseMap[namespace]; !exist {
```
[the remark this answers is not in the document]
junshun: So package controller installs packages one by one so this shouldn't be an issue. However we could look into synchronization

pkg/authenticator/ecrsecret_test.go
junshun: done

pkg/authenticator/ecrsecret.go
junshun: done by just initializing it to empty and filing it later as it gets updated anytime the config map is changed

pkg/authenticator/ecrsecret.go
junshun: Ahh this was really nice thanks

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	if update == "" {
+		delete(cm.Data, namespace)
+	} else {
+		cm.Data[namespace] = update
+	}
+}
+
+func (s *ecrSecret) GetSecretValues(ctx context.Context, namespace string) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
junshun: So this return matches the values struct that helm driver uses to pass in value. Are you suggesting to change it to return this struct only to have to change it back to map[string]interface to pass into the helmdriver later?

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+type ecrSecret struct {
+	clientset    kubernetes.Interface
+	nsReleaseMap map[string]string
+}
+
+var _ Authenticator = (*ecrSecret)(nil)
+
+func NewECRSecret(config *rest.Config) (*ecrSecret, error) {
```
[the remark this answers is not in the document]
junshun: Issue here is the config return by manager is not compatiable with rest.RESTClientFor due to missing GroupVersion.

pkg/authenticator/ecrsecret.go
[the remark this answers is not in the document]
ewollesen: If you really think it's that important, but I'd say it's a tomorrow problem, and you should deal with it in the way that makes the most sense... tomorrow. :) This being here is likely to confuse others, since it doesn't actually do anything now. It could also color your perceptions of the best solution to the problem tomorrow, when it's actually needed, that is, there might be a better solution to this tomorrow, but you wouldn't see it, because you've already made these decisions today.

pkg/authenticator/ecrsecret.go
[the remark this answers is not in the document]
ewollesen: We've been chatting about this in slack, but for documentation's sake, I'll repeat here. My understanding is that the groupversion should be set at instantiation, and it should match what the client is being used to retrieve/modify. In most cases for us, that means: eks-anywhere-packages/api/v1alpha1/groupversion_info.go:27

pkg/authenticator/ecrsecret.go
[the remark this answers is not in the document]
ewollesen: Yeah, I haven't seen anything to make me think that controllers operate in a multi threaded environment (unless we're the ones launching said threads) so I _think_ this is probably OK. Some of my go background was on a project that was highly concurrent, and this would have to have been wrapped with a mutex. I don't think it's a bad habit to wrap your maps in a mutex, but ... as I said, probably not necessary here.

pkg/authenticator/ecrsecret_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+)
+
+func TestAuthFilename(t *testing.T) {
+	config := rest.Config{}
+	t.Run("golden path for set HELM_REGISTRY_CONFIG", func(t *testing.T) {
+		testfile := "/test.txt"
+		t.Setenv("HELM_REGISTRY_CONFIG", testfile)
+		ecrAuth, _ := NewECRSecret(&config)
```
ewollesen, line 23: You don't get to skip checking errors in test files either! 😄 
```suggestion
		ecrAuth, err := NewECRSecret(&config)
		require.NoError(t, err)
```

pkg/driver/helmdriver.go
```diff
@@ -22,19 +22,18 @@ import (
...
 
-func NewHelm(log logr.Logger) (*helmDriver, error) {
+func NewHelm(log logr.Logger, secretAuth auth.Authenticator) (*helmDriver, error) {
 	settings := cli.New()
 
-	// TODO Catch error here if not provided docker config or continue without an authfile if possible
-	secretAuth := auth.NewHelmSecret()
-	authfile, _ := secretAuth.AuthFilename()
+	authfile := secretAuth.AuthFilename()
 	if authfile != "" {
```
ewollesen, line 37: Github won't let me suggest this via their web editor, cuz they're annoying, but...

Lines 37-39 are doing nothing. ClientOptCredentialsFile has no side-effects when called. It returns a ClientOpt (which itself is a function), and that ClientOpt should be passed to the registry.NewClient call on line 40. It should look more like this:

```
client, err := registry.NewClient(registry.ClientOptCredentialsFile(authfile))
```

Which all begs the question of if you really need to set the credentials file at all? As I assume this code was working before now..? I guess it got missed in previous reviews? Or am I just missing something?

I'm kind of suprised one of our many linters doesn't catch this.

pkg/driver/helmdriver.go
junshun: this was done in latest commit

pkg/driver/helmdriver.go
junshun: Done

pkg/driver/helmdriver.go
junshun: Done

pkg/driver/helmdriver.go
junshun: Moved out of helm driver

pkg/driver/helmdriver.go
junshun: fixed

pkg/driver/helmdriver.go
jonahjon: Ok

pkg/authenticator/ecrsecret.go
jonahjon: Sounds good

jonahjon (approved):

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,135 @@
...
+	var values []string
+	if data, ok := cm.Data[namespace]; ok {
+		values = strings.Split(data, ",")
+	}
+
+	if operation == ADD {
+		values = append(values, name)
+	}
+
+	if operation == ADD {
```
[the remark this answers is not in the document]
junshun: good catch. Fixed

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return err
+	}
+	// Store data
+	s.nsReleaseMap = cm.Data
+
+	return nil
+}
+
+func modifyToConfigMap(cm *v1.ConfigMap, name string, namespace string, operation int) {
```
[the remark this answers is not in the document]
junshun: Changed to addto and remove from from at the function above it to remove any confusion

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,155 @@
...
+		_, exists := check[val]
+		if !exists {
+			check[val] = true
+			result = append(result, val)
+		}
+	}
+
+	update := strings.Join(result, ",")
+	if update == "" {
+		delete(cm.Data, namespace)
```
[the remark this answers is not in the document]
junshun: fixed

pkg/authenticator/ecrsecret.go
```diff
@@ -0,0 +1,155 @@
...
+	values = append(values, name)
+	result := make([]string, 0, len(values))
+	check := map[string]bool{}
+	for _, val := range values {
+		_, exists := check[val]
+		if !exists {
+			check[val] = true
+			result = append(result, val)
+		}
+	}
```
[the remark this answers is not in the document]
junshun: This is to make sure we don't add duplicate names if there are multiple installs. Example lets say namespace a has 3 things installed  so example values would be ["a","b","c"]. If they were to install "a" again, we dont want to add it to values again.

pkg/driver/helmdriver_test.go
```diff
@@ -105,6 +109,15 @@ func TestIsConfigChanged(t *testing.T) {
...
 //
+func createNewHelm(t *testing.T) (error, *helmDriver) {
+	fakeRestClient := fakerest.RESTClient{
+		GroupVersion: api.GroupVersion,
+	}
+	secretAuth, err := auth.NewECRSecret(&fakeRestClient)
+	require.NoError(t, err)
+	helm, err := NewHelm(logr.Discard(), secretAuth)
+	return err, helm
+}
```
ewollesen, line 120:
```suggestion
func createNewHelm(t *testing.T) (*helmDriver,error) {
	fakeRestClient := fakerest.RESTClient{
		GroupVersion: api.GroupVersion,
	}
	secretAuth, err := auth.NewECRSecret(&fakeRestClient)
	require.NoError(t, err)
	helm, err := NewHelm(logr.Discard(), secretAuth)
	require.NoError(t, err)
	return helm, err
}
```
Putting the error before the helm driver is weird in golang speak.

junshun: changed and fixed tests to match

ewollesen (approved):
/lgtm

PR #336, opened 2022-08-02
Private-bundle support

jonahjon:

*Description of changes:*

Added support for writing bundles from ECR private.

In addition moved any functions dealing with GO SDK ecr or ecrpublic imagedetails, to shared function using a common data type. This is a step towards creating shared functions between ECR and ECRpublic, but didn't want the PR to be any bigger than needed so re-factored functions needed for this feature.


TerryHowe (approved):
/lgtm

PR #342, opened 2022-08-03
Remove unused struct from gen bundle

TerryHowe:
This thing wasn't being used and just noticed it in a search


vincentni (approved):
/lgtm

PR #343, opened 2022-08-03
Fix bundle states

TerryHowe:


ivyostosh:
I like the new states! Much cleaner than before. I'm just also wondering, how do you differentiate "ignored" with "invalid version"?

TerryHowe:
> I like the new states! Much cleaner than before. I'm just also wondering, how do you differentiate "ignored" with "invalid version"?


ignored is in an unmonitored namespace and invalid version has a name that we do not recognize

ivyostosh:
So is this a correct map from the previous states to the current states:
- inactive, active, active (upgrade available) -> available
- ignored -> ignored
- ignored version, invalid version -> invalid version

TerryHowe:
> So is this a correct map from the previous states to the current states:
> 
> * inactive, active, active (upgrade available) -> available
> * ignored -> ignored
> * ignored version, invalid version -> invalid version


ingored version gets mapped to available now. ingored version was for a version of k8s that did not match the server version

ivyostosh:
> * inactive, active, active (upgrade available) -> available
> * ignored -> ignored
> * ignored version, invalid version -> invalid version

So in summary - a map from the previous states to the current states:
inactive, active, active (upgrade available), ignored version -> available
ignored -> ignored
invalid version -> invalid version

PR #351, opened 2022-08-04
Adjustments to make sure Helm charts are public, images private

jonahjon:

*Description of changes:*

This is adjustments to make sure the helm charts remain public, while images remain private. It can be reverted back to the original once our helm driver supports pulling from private.


generatebundlefile/promote.go
```diff
@@ -151,17 +151,30 @@ func (c *SDKClients) PromoteHelmChart(repository, authFile string, crossAccount
...
 	// Unpack requires.yaml into a GO struct
-	helmRequires, err := validateHelmRequires(f)
+	helmRequiresImages, err := validateHelmRequires(f)
 	if err != nil {
 		return fmt.Errorf("Unable to parse requires.yaml file to Go Struct %s", err)
 	}
-	// Add the helm chart to the struct before looping through lookup/promote since we need it promoted too.
-	helmRequires.Spec.Images = append(helmRequires.Spec.Images, Image{Repository: chartName, Tag: version, Digest: sha})
+
+	// Create a 2nd struct since the the helm chart is going to Public ECR while the image are going to Private ECR.
```
ivyostosh, line 159:
```suggestion
	// Create a 2nd struct since the the helm chart is going to Public ECR while the images are going to Private ECR.
```

generatebundlefile/promote.go
```diff
@@ -207,6 +220,19 @@ func (c *SDKClients) PromoteHelmChart(repository, authFile string, crossAccount
...
+	// This will hard force the Helm chart to be moved from Private ECR to Public ECR if it didn't exist.
+	for _, images := range helmRequires.Spec.Images {
+		source := fmt.Sprintf("docker://%s.dkr.ecr.us-west-2.amazonaws.com/%s:%s", c.stsClient.AccountID, images.Repository, images.Tag)
+		destination := fmt.Sprintf("docker://%s/%s:%s", c.ecrPublicClient.SourceRegistry, images.Repository, images.Tag)
+		BundleLog.Info("Copying Helm Digest, and Tag to destination location......", "Location", fmt.Sprintf("%s/%s:%s %s", c.ecrPublicClient.SourceRegistry, images.Repository, images.Tag, images.Digest))
+		err = copyImage(BundleLog, authFile, source, destination)
+		if err != nil {
+			return fmt.Errorf("Unable to copy image from source to destination repo %s", err)
+		}
+		continue
```
ivyostosh, line 234: What does continue do here? Wouldn't it loop back naturally anyway?

generatebundlefile/promote.go
```diff
@@ -151,17 +151,30 @@ func (c *SDKClients) PromoteHelmChart(repository, authFile string, crossAccount
...
-	helmRequires, err := validateHelmRequires(f)
+	helmRequiresImages, err := validateHelmRequires(f)
 	if err != nil {
 		return fmt.Errorf("Unable to parse requires.yaml file to Go Struct %s", err)
 	}
-	// Add the helm chart to the struct before looping through lookup/promote since we need it promoted too.
-	helmRequires.Spec.Images = append(helmRequires.Spec.Images, Image{Repository: chartName, Tag: version, Digest: sha})
+
+	// Create a 2nd struct since the the helm chart is going to Public ECR while the image are going to Private ECR.
+	// Remove this once we add support for Private ECR pull in the Helm Driver.
```
junshun, line 160: I think you can get rid of this comment as long term when kubelet authentication is used we will need the helm charts in public ECR.

generatebundlefile/ecr_helper.go
```diff
@@ -180,3 +181,15 @@ func getLastestImageSha(details []ImageDetailsBothECR) (*api.SourceVersion, erro
 	}
 	return &api.SourceVersion{Name: latest.ImageTags[0], Digest: *latest.ImageDigest}, nil
 }
+
+// copyImagePrivPubSameAcct will copy an OCI artifact from ECR us-west-2 to ECR Public within the same account.
```
ivyostosh, line 185:
```suggestion
// copyImage will copy an OCI artifact from ECR us-west-2 to ECR Public within the same account.
```

ivyostosh (commented):
What about images that have been pushed to public ECR already?

generatebundlefile/promote.go
jonahjon: yea technically it's not needed here.

jonahjon: I'll take it out.

jonahjon: Good catch.

jonahjon:
Need CR-73735447 merged before we merge this.

jonahjon:
> What about images that have been pushed to public ECR already?

We can leave as is to be backwards compatible.

generatebundlefile/ecr_helper.go
```diff
@@ -180,3 +181,15 @@ func getLastestImageSha(details []ImageDetailsBothECR) (*api.SourceVersion, erro
 	}
 	return &api.SourceVersion{Name: latest.ImageTags[0], Digest: *latest.ImageDigest}, nil
 }
+
+// copyImage will copy an OCI artifact from ECR us-west-2 to ECR Public within the same account.
```
ewollesen, line 185: I wonder about the veracity of this comment, given that "source" and "destination" could be any two (possibly different?) OCI compliant registries, for which "authfile" has access, right?

ewollesen (commented):
/lgtm

PR #352, opened 2022-08-04
Adding ecrtokenrefresher to gosec exclude dir

junshun:
*Description of changes:*
Adding ecrtokenrefresher to gosec exclude dir as gosec can only handle one module per repository.


TerryHowe (approved):
/lgtm
/approve

It would be nice if we had another gosec run for those subdirs

PR #353, opened 2022-08-04
Private registry

TerryHowe:
Currently, we use the registry for the bundle to install every package so that we can support private registries. This works for now since we currently only support everything from public ecr or a private registry. This change allows us to support:

1. private registries
2. sourceRegistry specified in the package custom resource
3. registry specified in the bundle
4. default registry will be the source registry of the bundle if nothing else is specified.

### DIFF
--- a/api/v1alpha1/packagebundlecontroller_types.go
+++ b/api/v1alpha1/packagebundlecontroller_types.go
@@ -64,6 +64,10 @@ type PackageBundleControllerSpec struct {
 	// +optional
 	ActiveBundle string `json:"activeBundle"`
 
+	// PrivateRegistry is the registry being used for all images, charts and bundles
+	// +optional
+	PrivateRegistry string `json:"privateRegistry"`
+
 	// +kubebuilder:validation:Required
 	// Source of the bundle.
 	Source PackageBundleControllerSource `json:"source"`
--- a/charts/eks-anywhere-packages/crds/crd.yaml
+++ b/charts/eks-anywhere-packages/crds/crd.yaml
@@ -27,34 +27,45 @@ spec:
     name: v1alpha1
     schema:
       openAPIV3Schema:
-        description: PackageBundleController is the Schema for the packagebundlecontrollers API
+        description: PackageBundleController is the Schema for the packagebundlecontrollers
+          API.
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
           metadata:
             type: object
           spec:
-            description: PackageBundleControllerSpec defines the desired state of PackageBundleController
+            description: PackageBundleControllerSpec defines the desired state of
+              PackageBundleController.
             properties:
               activeBundle:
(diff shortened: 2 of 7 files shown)


api/v1alpha1/packagebundlecontroller_types.go
```diff
@@ -64,6 +64,10 @@ type PackageBundleControllerSpec struct {
 	// +optional
 	ActiveBundle string `json:"activeBundle"`
 
+	// PrivateRegistry is the OCR address hosting the bundle.
```
lewisdiamond, line 67: OCI + ECR = OCR?

pkg/packages/manager.go
```diff
@@ -60,7 +61,9 @@ func processInstalling(mc *ManagerContext) bool {
 		mc.Log.Error(err, "Install failed")
 		return true
 	}
-	values[sourceRegistry] = mc.Source.Registry
+	if mc.Registry != "" && values[sourceRegistry] == "" {
```
junshun, line 64: Is there a scenario where both of these are set? I would image if Registry is set the user would want to use that regardless of sourceRegistry being there?

charts/eks-anywhere-packages/crds/crd.yaml
```diff
@@ -210,6 +214,10 @@ spec:
                               schema:
                                 description: Schema is a base64 encoded, gzipped json schema used to validate package configurations
                                 type: string
+                              schema:
```
lewisdiamond, line 217: What's with this?

charts/eks-anywhere-packages/crds/crd.yaml
```diff
@@ -300,6 +308,10 @@ spec:
                                   schema:
                                     description: Schema is a base64 encoded, gzipped json schema used to validate package configurations
                                     type: string
+                                  schema:
```
lewisdiamond, line 311: And this?

charts/eks-anywhere-packages/crds/crd.yaml
```diff
@@ -321,7 +333,7 @@ spec:
                 enum:
                 - available
                 - ignored
-                - invalid version
```
lewisdiamond, line 324: Wasn't that already merged?

lewisdiamond:
Can you explain what this feature is for?

api/v1alpha1/packagebundlecontroller_types.go
TerryHowe: copy pasta error from PackageBundleControllerSource

charts/eks-anywhere-packages/crds/crd.yaml
TerryHowe: no idea why it generated that

pkg/packages/manager.go
TerryHowe: So this `values[sourceRegistry]` comes from Package and that should override everything

junshun: So does that mean sourceRegistry from PackageBundleController spec does nothing?

charts/eks-anywhere-packages/crds/crd.yaml
TerryHowe: uggh, missed that.

api/v1alpha1/packagebundlecontroller_types.go
TerryHowe: fixed

charts/eks-anywhere-packages/crds/crd.yaml
TerryHowe: fixed

charts/eks-anywhere-packages/crds/crd.yaml
TerryHowe: fixed

charts/eks-anywhere-packages/crds/crd.yaml
TerryHowe: fixed

pkg/packages/manager.go
TerryHowe: What should happen is
1. use `values[sourceRegistry]` which is from Package
2. use privateRegistry from PackageBundleController
3. use registry from bundle
4. use bundle registry from PackageBundleController

ewollesen: Is there a good way to get the comment above into the code in some (preferably) obvious and readable way?

TerryHowe: Good idea, added a method to ManagerContext

TerryHowe:
> Can you explain what this feature is for?

Yes, updated the description

pkg/packages/manager.go
TerryHowe: added some tests too

lewisdiamond:
This is missing the helm CRDs update. Run `make helm-build`.

jonahjon:
Does this mean we want to add back Bundle the field `api.BundlePackageSource.Registry`?

TerryHowe:
> Does this mean we want to add back Bundle the field `api.BundlePackageSource.Registry`? Or should we continue to remove it from the bundle as we have been.

We should leave it out.

TerryHowe:
> This is missing the helm CRDs update. Run `make helm-build`.

done

PR #354, opened 2022-08-04
Remove export test

TerryHowe:
One of these wasn't used, the other it would be easier to make it public.

### DIFF
--- a/api/v1alpha1/export_test.go
+++ b/api/v1alpha1/export_test.go
@@ -1,5 +1,0 @@
-// Export private functions and methods for testing
-package v1alpha1
-
-var GetMajorMinorBuild = (*PackageBundle).getMajorMinorBuild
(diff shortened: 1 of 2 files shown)


ewollesen:
I'm against exporting a function solely so it can be tested from another package.

Non-exported methods should not be tested from other packages.

Making an non-exported method exported, simply so you can test it, is dangerous. Once exported, we can no longer change this function without the risk of breaking others. We should be careful about what we export, and not do it just to aid an artificial barrier in our testing.

TerryHowe:
> I'm against exporting a function solely so it can be tested from another package.

Okay, but I just replaced one export for another.

TerryHowe:
Just put the test in the same package

ewollesen (approved):
/lgtm

PR #361, opened 2022-08-05
Setup envtest

TerryHowe:
This avoids the expectation that GOBIN is in your PATH


TerryHowe:
github actions is borked

PR #374, opened 2022-08-08
Active bundle matches k8s version

TerryHowe:
Validate activeBundle is right kubeVersion

### DIFF
--- a/pkg/webhook/packagebundlecontroller_webhook.go
+++ b/pkg/webhook/packagebundlecontroller_webhook.go
@@ -17,18 +17,19 @@ package webhook
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/webhook/packagebundlecontroller_webhook.go
```diff
@@ -0,0 +1,157 @@
...
+					Status:  metav1.StatusFailure,
+					Code:    http.StatusNotFound,
+					Message: msg,
+					Reason:  metav1.StatusReasonNotFound,
+				},
+			},
+		}
+		return resp, nil
+	}
+
```
TerryHowe, line 109: This is the new stuff here. For some reason github is showing this as a new rather than move/change

PR #380, opened 2022-08-09
Asserts

TerryHowe:
Mindless work for fly time

### DIFF
--- a/api/v1alpha1/package_test.go
+++ b/api/v1alpha1/package_test.go
@@ -46,6 +46,6 @@ notactuallyyaml
     `
 	ao := givenPackage(provided)
 	_, err := ao.GetValues()
-	assert.EqualError(t, err, "asdf")
+	assert.EqualError(t, err, "error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {}")
 	assert.Contains(t, err.Error(), "error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {}")
 }
--- a/api/v1alpha1/package_test.go
+++ b/api/v1alpha1/package_test.go
@@ -46,6 +46,6 @@ notactuallyyaml
     `
 	ao := givenPackage(provided)
 	_, err := ao.GetValues()
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "error unmarshaling")
+	assert.EqualError(t, err, "asdf")
+	assert.Contains(t, err.Error(), "error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {}")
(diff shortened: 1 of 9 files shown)


api/v1alpha1/package_test.go
```diff
@@ -46,6 +46,6 @@ notactuallyyaml
     `
 	ao := givenPackage(provided)
 	_, err := ao.GetValues()
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "error unmarshaling")
+	assert.EqualError(t, err, "error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {}")
+	assert.Contains(t, err.Error(), "error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {}")
```
ewollesen, line 50: I don't think both of these are necessary. Were it me, I'd keep the EqualError on 49, and delete line 50, but it also doesn't hurt anything I guess. But they're pretty much literally doing the same thing.

controllers/package_controller_test.go
```diff
@@ -135,14 +129,10 @@ func TestReconcile(t *testing.T) {
 		sut := tf.newReconciler()
 		req := tf.mockRequest()
 		got, err := sut.Reconcile(ctx, req)
-		if err != nil {
-			t.Errorf("expected <nil> got <%s>", err)
-		}
+		assert.Nil(t, err)
```
ewollesen, line 132: This works for sure, but for future reference. `assert.NoError(t, err)` may better communicate your intentions.

pkg/bundle/registry_client_test.go
```diff
@@ -70,9 +54,7 @@ func TestDownloadBundle(t *testing.T) {
 		ref := fmt.Sprintf("%s:%s-%s", baseRef, kubeVersion, tag)
 
 		_, err := ecr.DownloadBundle(ctx, ref)
-		if err == nil {
-			t.Errorf("expected error, got nil")
-		}
+		assert.EqualError(t, err, "pulling package bundle: test error")
```
ewollesen, line 57: There exists [`assert.AnError`](https://pkg.go.dev/github.com/stretchr/testify/assert#pkg-variables) which can expression this kind of intention very well, plus it would let you use `assert.ErrorIs(t, assert.AnError)`, which is less brittle, as it doesn't depend on the text within the error.

ewollesen (approved):
Just some food for thought comments, no need to change anything.

/lgtm

pkg/bundle/registry_client_test.go
TerryHowe: Yeh, it was easier to just dump the full text in. It would be nice if there was something like `assertErrorContains` to get a substring of the error

PR #390, opened 2022-08-11
Tag length check failing release script

jonahjon:

*Description of changes:*

Failure on lookup from ECR where some images get tags overridden, and have a tag length of 0.


TerryHowe (approved):
/lgtm

PR #391, opened 2022-08-11
Rename PrivateRegistry to LocalRegistry

TerryHowe:
The Name LocalRegistry is perhaps less confusing than PrivateRegistry


TerryHowe:
/hold

Discuss for now

PR #396, opened 2022-08-11
Trying another fix for release

jonahjon:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/generatebundlefile/hack/release.sh
+++ b/generatebundlefile/hack/release.sh
@@ -30,7 +30,7 @@ EOF
 # Release Package Images to Packages Artifact account
 BASE_DIRECTORY=$(git rev-parse --show-toplevel)
 export AWS_CONFIG_FILE=${BASE_DIRECTORY}/generatebundlefile/configfile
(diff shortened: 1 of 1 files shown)


generatebundlefile/hack/release.sh
```diff
@@ -30,7 +30,7 @@ EOF
 # Release Package Images to Packages Artifact account
 BASE_DIRECTORY=$(git rev-parse --show-toplevel)
 export AWS_CONFIG_FILE=${BASE_DIRECTORY}/generatebundlefile/configfile
```
TerryHowe, line 32: I don't think that export is going to do anything, but I could be missing something

TerryHowe (approved):
/lgtm

generatebundlefile/hack/release.sh
jonahjon: It is a supported field for the CLI https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

PR #398, opened 2022-08-12
generatebundlefile: update oras to 0.13.0.

ewollesen:
This upgrades our oras cli to 0.13.0 which adds the --password-stdin flag to the push and pull commands, which in turn allows us to simplify our authentication with oras.


jonahjon:
I like it!
probably merge this after GA in case it breaks something though.

TerryHowe (approved):
/lgtm
/approve

PR #403, opened 2022-08-15
Group version fix for RestAPI CoreV1

junshun:
Fixed RestClient Group Version for CoreV1 API

### DIFF
--- a/controllers/package_controller.go
+++ b/controllers/package_controller.go
@@ -44,6 +44,9 @@ import (
 const (
 	packageName = "Package"
 	retryLong   = time.Second * time.Duration(60)
+	groupName   = ""
+	groupVer    = "v1"
+	apiPath     = "/api"
 )
 
 // PackageReconciler reconciles a Package object
@@ -115,11 +118,11 @@ func RegisterPackageReconciler(mgr ctrl.Manager) (err error) {
 }
 
 func createSecretAuth(cfg *rest.Config) (auth.Authenticator, error) {
-	gv := schema.GroupVersion{Group: "", Version: "v1"}
+	gv := schema.GroupVersion{Group: groupName, Version: groupVer}
 	cfg.GroupVersion = &gv
 	cfg.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
 	cfg.UserAgent = rest.DefaultKubernetesUserAgent()
-	cfg.APIPath = "/api"
+	cfg.APIPath = apiPath
 	restclient, err := rest.RESTClientFor(cfg)
 	if err != nil {
 		return nil, fmt.Errorf("creating rest client from config %s", err)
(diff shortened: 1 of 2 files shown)


controllers/package_controller.go
```diff
@@ -114,9 +115,11 @@ func RegisterPackageReconciler(mgr ctrl.Manager) (err error) {
 }
 
 func createSecretAuth(cfg *rest.Config) (auth.Authenticator, error) {
-	scheme := runtime.NewScheme()
-	cfg.GroupVersion = &api.GroupVersion
-	cfg.NegotiatedSerializer = serializer.NewCodecFactory(scheme)
```
TerryHowe, line 119: This new approach has all sorts of hard coded strings. No better way?

junshun: fixed

ewollesen: I think the constants are OK, but FWIW, the k8s authors seem to be OK with using string literals in the instantiation of a GroupVersion. For example:
```sh
$ grep -r 'GroupVersion{Group: "", Version: "v1"}' | wc -l
914
```

ewollesen: You might consider copying the cfg in this function too... It's kinda weird that a function named "createSecretAuth" would modify the argument you give it.

TerryHowe: Whenever you start talking about the principle of least surprise, I think of the law of surprise from the Witcher. No first born children should b exchanged here.

I can live with the hard coded strings, I'm just a little surprised they don't have something better.

ewollesen: > Whenever you start talking about the principle of least surprise, I think of the law of surprise from the Witcher. No first born children should b exchanged here.
> 
> I can live with the hard coded strings, I'm just a little surprised they don't have something better.

LOL. I love the Witcher! :) (And no one wants my first-born, I promise you! :P) 

I was surprised too, I thought they'd have some exported var ready to use or something... and to be clear, I really don't mean this as a counter to your suggestion Terry, just I noticed that Steven was following the pattern of the authors.

ewollesen: But now I have to ask, did you find any surprises when you returned home from Seattle?

pkg/csset/csset_test.go
```diff
@@ -29,7 +30,9 @@ func TestCSSetTwo(t *testing.T) {
 	set := NewCSSet("")
 	set.Add("one")
 	set.Add("two")
-	assert.Equal(t, "one,two", set.String())
+	exp := [2]string{"one", "two"}
```
ewollesen, line 33: If we end up using csset anywhere else, maybe we should write a test helper for this. Something that blends with the testify/assert package would be nice.

controllers/package_controller.go
TerryHowe: Ha! No surprises for me.

PR #405, opened 2022-08-15
Comply with higher level linting

TerryHowe:
The `make lint` target complains about these minor issues. I'd rather not see it.


ewollesen:
A perfect example of why linters are terrible.

/lgtm

PR #408, opened 2022-08-15
Downgrade oras version due to ECR bug.

jonahjon:

*Issue #, if available:*

This is issue actively blocking pipelines.

*Description of changes:*

Oras 13 is not working correctly, and blocking pipeline. 

Link to [issue](https://github.com/oras-project/oras/issues/456#issuecomment-1195531498
), and comment from maintainers.

```
As a registry client, ORAS should do the layer pushing in two step:

Try to initiate an upload process via a POST request
Do the blob upload via PUT or PATCHes
During PUT(step 2), ORAS fails to reuse the token obtained by POST(step 1) and ends in rewinding the request body.

Why it only fails for ECR:
An credential cache was introduced in oras-go v2. The cache was keyed by the scopes in the oauth challenge. An OCI-compliant format of the scope should be like repository:some-project/some-repo:pull,push. ECR returns aws and that unexpected key prevents PUT(step 2) from getting the token cached by POST(step 1).

We had a fix in oras-go and will apply the change to CLI. This issue will be fixed in upcoming ORAS release.

cc @nima
```

### DIFF
--- a/generatebundlefile/hack/generate_bundle.sh
+++ b/generatebundlefile/hack/generate_bundle.sh
@@ -46,10 +46,8 @@ function generate () {
 function push () {
     local version=$1
     cd "${BASE_DIRECTORY}/generatebundlefile/output"
-    awsAuth "ecr-public" | "$ORAS_BIN" push --password-stdin \
-        "${REPO}:v${version}-${CODEBUILD_BUILD_NUMBER}" bundle.yaml
-    awsAuth "ecr-public" | "$ORAS_BIN" push --password-stdin \
-        "${REPO}:v${version}-latest" bundle.yaml
+    awsAuth "ecr-public" | "$ORAS_BIN" push "${REPO}:v${version}-${CODEBUILD_BUILD_NUMBER}" bundle.yaml
+    awsAuth "ecr-public" | "$ORAS_BIN" push "${REPO}:v${version}-latest" bundle.yaml
 }
 
 for version in 1-21 1-22; do
--- a/generatebundlefile/Makefile
+++ b/generatebundlefile/Makefile
@@ -61,7 +61,7 @@ test:
 presubmit: build test # lint is run via github action
 
 ## ORAS Make commands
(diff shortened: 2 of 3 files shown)


generatebundlefile/hack/generate_bundle.sh
```diff
@@ -46,10 +46,8 @@ function generate () {
 function push () {
     local version=$1
     cd "${BASE_DIRECTORY}/generatebundlefile/output"
-    awsAuth "ecr-public" | "$ORAS_BIN" push --password-stdin \
-        "${REPO}:v${version}-${CODEBUILD_BUILD_NUMBER}" bundle.yaml
-    awsAuth "ecr-public" | "$ORAS_BIN" push --password-stdin \
-        "${REPO}:v${version}-latest" bundle.yaml
+    awsAuth "ecr-public" | "$ORAS_BIN" push "${REPO}:v${version}-${CODEBUILD_BUILD_NUMBER}" bundle.yaml
+    awsAuth "ecr-public" | "$ORAS_BIN" push "${REPO}:v${version}-latest" bundle.yaml
```
TerryHowe, line 50:
```suggestion
    awsAuth "ecr-public"
    "$ORAS_BIN" push "${REPO}:v${version}-${CODEBUILD_BUILD_NUMBER}" bundle.yaml
    "$ORAS_BIN" push "${REPO}:v${version}-latest" bundle.yaml
```

TerryHowe: That is how it used to be. I was just wondering why you were piping the output of that to `oras push`. I think what you have will work fine, it is just a bit odd.

TerryHowe (commented):
/lgtm
/approve

generatebundlefile/hack/generate_bundle.sh
jonahjon: I think it's because of the changes Eric made to the `common.sh` it required being piped now.

PR #415, opened 2022-08-17
Fix issue where config was detected as change for imagePullSecrets

junshun:
Address package config changing due to imagePullSecrets being added by helmdriver and thus not matching the original passed in config.


TerryHowe (approved):
/lgtm
/approve

PR #422, opened 2022-08-17
Reduce the time waiting for the webhook

TerryHowe:


TerryHowe:
/hold

for discussion right now

jonahjon:
It's a bit brittle, but I'm ok with this if we file an issue to investigate a better solution post GA.

PR #423, opened 2022-08-17
Switching to Warning Logger for Refresher when missing namespace

junshun:
Switching to warning instead of error in print message when missing namespace (mostly due to emissary-system)


TerryHowe (approved):
/lgtm
/approve

PR #428, opened 2022-08-18
package bundle controller: requeue quickly if webhook offline

ewollesen:
Upon cluster startup, the package bundle controller finds itself without a
bundle. So it reaches out the registry, and tries to create and activate the
newest bundle the registry has. If the package bundle validating webhook has
not yet come online (indicated by a connection refused error), then this code
will requeue a reconcile request for just a short time, so things aren't held
up for hours.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/controllers/packagebundlecontroller_controller.go
+++ b/controllers/packagebundlecontroller_controller.go
@@ -32,8 +32,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


ewollesen:
/hold

I need to do a final test or two.

TerryHowe (approved):
/lgtm

PR #439, opened 2022-08-24
Change the default make target

TerryHowe:
Once you modify a file, the make never works again now. Change it so the default target is just like presubmit, but doesn't expect no git diffs.


vincentni (approved):
/approve
/lgtm

PR #440, opened 2022-08-25
Defaulting region to us-west-2 and exit gracefully 

junshun:
This should prevent job pod from Crash looping


TerryHowe (approved):
/lgtm
/approve

PR #441, opened 2022-08-26
Make test utils consistent

TerryHowe:
There are I think 4 different ways to make these things, so this removes one.


vincentni (approved):
/approve
/lgtm

PR #445, opened 2022-08-28
Remove unused fake manager

TerryHowe:
I started looking at this fake manager and it turns out we just weren't using it either.


vincentni (approved):
/lgtm
/approve

PR #446, opened 2022-08-29
Consolidate testutil

TerryHowe:


pkg/file/filereader_test.go
```diff
@@ -64,18 +64,3 @@ func TestFileReaderParseGood(t *testing.T) {
 
 	assert.Nil(t, actual)
 }
-
```
TerryHowe, line 67: This was causing a circular dependency and these test are tested elsewhere. Long range, I think we should try to get rid of these files as much as we can

PR #447, opened 2022-08-29
Use mock puller for consistency

TerryHowe:
We use gomock everywhere except here.


ewollesen:
I think this is the wrong call. Our usage of gomock makes our tests brittle, and over-complicates simple tests and simple changes. It leads to mocking out the detailed implementations of our functions, rather than their external behaviors. A lot of this may be due to our usage of gomock, but it remains an issue all the same. I would rarely, if ever, choose gomock. But rest assured, this is the last you'll hear from me about it. :)

TerryHowe:
I disagree for several reasons:
* everything else uses gomock and there is value in consistency
* you call it brittle, I call it precise Use `gomock.Any()` if you want. That is completely fine with me.
* gomock makes the test more understandable because the test specifies the returns, not some code elsewhere
* gomock style is very consistent with just about every other languages mocks

PR #449, opened 2022-08-30
e2e docs overview

ewollesen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/testing/e2e-plan.md
+++ b/docs/testing/e2e-plan.md
@@ -1,0 +1,79 @@
+# Curated Packages End-To-End Testing Plan
+
+This document summarizes the current state and future plans for end-to-end (E2E) testing in the EKS-A Curated Packages (CP) project.
+
+## The Present
+
+End-to-end tests can be written to exercise CP workflows. Several parameters can be selected to control the environment (cluster) in which a test runs. These include: EKS-A provider, cluster node OS, and Kubernetes (K8s) version. For examples, see [TestCPackagesHarborInstallSimpleFlow](https://github.com/aws/eks-anywhere/blob/de84797d702028e15deb3363bbe222489f00229d/test/e2e/harbor_test.go) and [TestCPackagesVSphereKubernetes121SimpleFlow](https://github.com/aws/eks-anywhere/blob/de84797d702028e15deb3363bbe222489f00229d/test/e2e/curated_packages_test.go#L102).
+
+Tests have been written that cover the installation and running of the CP packages controller and the [hello-eks-anywhere test application](https://github.com/aws/eks-anywhere-build-tooling/blob/61628f1669044a39ff561385c7032e3b6c7d12ed/projects/aws-containers/hello-eks-anywhere/conf/hello.sh) across the three dimensions mentioned above (EKS-A provider, node OS, and K8s version). Additional tests have been written to exercise Harbor and MetalLB package installations. The results of these tests are reported in the appropriate EKS-A communications channels.
+
+*But the existing tests are not nearly enough to provide confidence that our packages are bug-free.* The current CP E2E framework does not (without significant effort by the test writer) run the existing tests against the versions of their artifacts that matter most[^1]. Nor do they run them for each submitted pull request (PR). These things are not the fault of the tests themselves, nor the tests' writers, but rather they've exposed the limitations of the current CP E2E framework.
+
+[^1]: Early E2E tests were run against production artifacts, generally because they were the only reliably accessible artifacts at the time.
+
+## The End-Goal
+
+To be able to provide confidence that the CP offerings are working across our supported EKS-A versions, EKS-A providers, cluster node OSes, and K8s versions, the current framework must be improved in the following ways:
+
++ **Artifact version control:**
+  Tests should be run against an arbitrary versions of artifacts, defaulting to their latest versions.
+
++ **Timing:**
+  E2E tests should be run on every revision of an "ok to test" PR. This may present some challenges for the current E2E test architecture, as it might require communication between separate AWS pipelines, or failing that, custom code that can communicate with other AWS resources to identify and list available artifacts built in other pipelines.
+
++ **Platform configuration:**
+  A test should be able to exclude specific EKS-A versions, EKS-A providers, cluster node OSes, or K8s versions with which they are incompatible or unsupported. A test will be run on every permutation by default.
+
++ **Configuration review & control:**
+  The test dimensions should be easily viewed, verified, and edited by humans (or reasonable facsimiles thereof). This provides a means of quickly excluding tests categorically for reasons such as hardware outage.
+
+Additionally, it is foreseen that the number of E2E tests that need to be run for each PR will quickly grow, and so it is recommended that these performance goals also be considered:
+
++ **Reusable clusters:** The amount of time required to create and destroy clusters before and after each test will quickly become overwhelming with reasonable amounts of hardware.
+
+## The Path Forward
+
+### Short Term
+
+#### Test Against Current Branch
+
+The current tests will be modified to run against the latest versions of artifacts. This will be accomplished by removing the existing hardcoded registry links from the existing tests, and implementing a means by which URLs to test-run-specific artifacts will be used. These up-to-date artifacts will be used (at a minimum) when installing the EKS-A Curated Packages controller during cluster creation, as well as when said controller acquires its active bundle. The newly-acquired active bundle will likewise contain links to newly-built test artifacts, such that the end-to-end tests will be run against artifacts built using the current git commits under test, and not the hard-coded development or production registries, as is the case now.
+
+Test runs should be accomplished using development ECR repositories (i.e. public ECR), but if or where that isn't the case, credentials for accessing a private ECR would have to be acquired, which could prove a technical risk, as to my knowledge, this isn't presently done.
+
+#### Investigate Daily E2E Run
+
+Running end-to-end tests more frequently is another priority of the short term plan. An investigation into the feasibility of adding a periodic run of the end-to-end tests will be performed. If a periodic end-to-end run is estimated to be within a week's work, then it will be undertaken. The intention is to be notified of end-to-end test failures before a release branch is created, when deadlines are already set and there are possibly many time-sensitive tasks already waiting to be completed.
+
+### Long Term
+
+Following the short term goals, several broader, more in-depth changes will help our end-to-end tests further raise our confidence. What follows is a brief summary of the larger components or actions required, with a more detailed design after the short term priorities have been satisfied.
+
+#### Test Manifests
+
+A list of each artifact produced, along with identifiers and URLs to acquire each artifact for testing. This manifest will serve as one input to a new testing framework that will be able to ensure that the desired combinations of providers, host OSes, and any other test dimensions deemed appropriate are run.
+
+This manifest will be generated in response to a pull request change or commit to selected branches (main, release-x, etc) of the eks-anywhere-packages, eks-anywhere, and eks-anywhere-build-tooling repositories (the last of which possibly limited to changes affecting a curated packages project), and will in turn be used by the testing framework.
+
+#### Increased Frequency Of E2E Test Runs
+
+If feasible, end-to-end tests will be run on each (ok-to-test) pull request on the eks-anywhere-packages repository. If that isn't feasible, then hourly or semi-hourly test runs will help catch problems sooner.
+
+Aside: The quality of curated packages relies upon the quality of our end-to-end tests. Correspondingly, the more often we run them, the faster we learn of, and can fix, problems. If we're not able to run the full battery of tests for each PR, then it is strongly recommended that we do whatever is in our power to be able to do so.
+
+#### Test Configuration Control
+
+A means by which tests can be disabled or enabled based upon their properties, requirements, or current statuses. In plainer words, a way to write a test that doesn't run for a specific provider, or to temporarily disable tests for specific host OS. This input, along with the test manifest are envisioned as being the bulk of the input to a given CP E2E test run.
+
+#### End-To-End Test Documentation
+
+Documentation will be generated that, in addition to the forthcoming design document, will aid developers and other contributors understand the end-to-end test framework, specifically how to write a new end-to-end test, how to run an end-to-end test outside of the build pipelines, how to be notified of failures, and how to debug test failures.
+
+A stretch goal could be to add documentation-style tests in the spirit of [Rust's documentation tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) or [go test examples](https://pkg.go.dev/testing#hdr-Examples).
+
+#### Independent Build Tooling and Testing
+
+In order for new CP end-to-end tests to be as straightforward as possible, and to run with the desired frequency goals, it is necessary to separate our build-tooling processes from those currently existing in the eks-anywhere and eks-anywhere-build-tooling repositories. The current processes produce significant developer friction when writing tests as well as running tests outside of the build pipelines. The existing system's builds are not as finely-grained as is required to meet the test-run-frequency goals defined previously in this document. Lastly, because the concerns of CP tests have little overlap with those of the EKS-A CLI, it is deemed appropriate that these distinctions in build and testing frameworks would allow each project to move more freely in directions that serves their individual goals best, while any infrastruture or utilities common to both could live in a neutral repository and shared by each.
+
+There will likely be some small number of the current CP E2E tests, particularly those concerned with verifying that a CLI installation of the EKS-A CP controller is successful, that could remain in the eks-anywhere repository. The existing Prow infrastructure will be used, and reporting will be delivered in the same methods to minimize disruption to other systems, habits, or customs previously developed.


docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,48 @@
+# Curated Packages End-To-End Testing Plan
+
+This document summarizes the current state and future plans for end-to-end (E2E) testing in the EKS-A Curated Packages (CP) project.
```
lewisdiamond, line 3: (Personal opinion) I prefer just "packages" instead of "CP".

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,48 @@
...
+
+## The Present
+
+End-to-end tests can be written to exercise CP workflows. Several parameters can be selected to control the environment (cluster) in which a test runs. These include: EKS-A provider, node OS, and Kubernetes (K8s) version. For examples, see [TestCPackagesHarborInstallSimpleFlow](https://github.com/aws/eks-anywhere/blob/de84797d702028e15deb3363bbe222489f00229d/test/e2e/harbor_test.go) and [TestCPackagesVSphereKubernetes121SimpleFlow](https://github.com/aws/eks-anywhere/blob/de84797d702028e15deb3363bbe222489f00229d/test/e2e/curated_packages_test.go#L102).
```
lewisdiamond, line 7: Let's change the "These include" for an actual list and let's try to make it as exhaustive as possible. I'd add `eks-anywhere version`.

ewollesen: Ah, well, see, this was meant to summarize the current state, and in the current state, the "framework" as it were doesn't support different eks-a versions. I'm not sure, but i think it would take a lot hoop-jumping and manual installation. This is a side-effect of the packages e2e tests running from the eks-a repo.

docs/testing/e2e-plan.md
ewollesen: I think I'm going to try sticking with CP, because it's more explicit, and shorter. Package feels like such an overloaded term to begin with, and I don't think any of us will stop using it, but the explicitness of CP feels clearer for documentation.

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,73 @@
...
+To be able to provide confidence that the CP offerings are working across our supported providers, OSes, and K8s versions, the current framework must be improved in the following ways:
+
++ **Artifact version control:**
+  Tests should be run against the latest versions of artifacts. [??Bonus points for letting test writers specify specific versions.??]
+
++ **Timing:**
+  E2E tests should be run on every revision of an "ok to test" PR.
+
++ **Platform configuration:**
+  A test should be able to exclude specific providers, OSes, or K8s versions with which they are incompatible or unsupported.
```
lewisdiamond, line 24: Maybe mention that it would otherwise run on all default combinations unless explicitly excluded

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,73 @@
...
+
+If feasible, end-to-end tests will be run on each (ok-to-test) pull request on the eks-anywhere-packages repository. If that isn't feasible, then hourly or semi-hourly test runs will help catch problems sooner.
+
+#### Test Configuration Control
+
+A means by which tests can be disabled or enabled based upon their properties, requirements, or current statuses. In plainer words, a way to write a test that doesn't run for a specific provider, or to temporarily disable tests for specific host OS. This input, along with the test manifest are envisioned as being the bulk of the input to a given CP E2E test run.
+
+#### End-To-End Test Documentation
+
+Documentation will be generated that, in addition to the forthcoming design document, will aid developers and other contributors understand the end-to-end test framework, specifically how to write a new end-to-end test, how to run an end-to-end test outside of the build pipelines, how to be notified of failures, and how to debug test failures.
```
lewisdiamond, line 67: I'd add a stretch goal of adding documentation tests. Similar to https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

docs/testing/e2e-testing-framework.md
```diff
@@ -0,0 +1,7 @@
+# Curated Packages End-To-End Testing Framework
```
lewisdiamond, line 1: Is this file still needed?

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,73 @@
...
+
+Tests have been written that cover the installation and running of the CP packages controller and the [hello-eks-anywhere test application](https://github.com/aws/eks-anywhere-build-tooling/blob/61628f1669044a39ff561385c7032e3b6c7d12ed/projects/aws-containers/hello-eks-anywhere/conf/hello.sh) across the three dimensions mentioned above (EKS-A provider, node OS, and K8s version). Additional tests have been written to exercise Harbor and MetalLB package installations. The results of these tests are reported in the appropriate EKS-A communications channels.
+
+*But the existing tests are not nearly enough to provide confidence that our packages are bug-free.* The current CP E2E framework does not (without significant effort by the test writer) run the existing tests against the versions of their artifacts that matter most. Nor do they run them for each submitted pull request (PR). These things are not the fault of the tests themselves, nor the tests' writers, but rather they've exposed the limitations of the current CP E2E framework.
```
lewisdiamond, line 11: Maybe add references and explanations on why the wrong thing is tested today.

docs/testing/e2e-plan.md
ewollesen: 👍🏻 Will do.

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,73 @@
...
+
+Following the short term goals, several broader, more in-depth changes will help our end-to-end tests further raise our confidence. What follows is a brief summary of the larger components or actions required.
+
+A more detailed design will be submitted for review after the short term priorities have been satisfied.
+
+#### Test Manifests
+
+A list of each artifact produced, along with identifiers and URLs to acquire each artifact for testing. This manifest will serve as one input to a new testing framework that will be able to ensure that the desired combinations of providers, host OSes, and any other test dimensions deemed appropriate are run.
+
+This manifest will be generated from a commit to the eks-anywhere-packages repository, and will in turn be used by the testing framework.
```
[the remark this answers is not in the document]
ewollesen: > the manifest should be created for any commit of any packages

I'm taking this to mean a change in any of the curated packages projects within the eks-anywhere repo...? If so, yes, I'll add that.

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### Increased Frequency Of E2E Test Runs
+
+If feasible, end-to-end tests will be run on each (ok-to-test) pull request on the eks-anywhere-packages repository. If that isn't feasible, then hourly or semi-hourly test runs will help catch problems sooner.
```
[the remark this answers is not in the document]
ewollesen: Will do.

docs/testing/e2e-testing-framework.md
ewollesen: Nah, it was just a placeholder. I'll nuke it.

docs/testing/e2e-plan.md
ewollesen: I've added a footnote with an explanation. I'm not sure what you would want for a reference? A link to the existing code? I'm going to assume this is good unless you say otherwise (which you're welcome to do).

docs/testing/e2e-plan.md
ewollesen: ...or? https://pkg.go.dev/testing#hdr-Examples

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
++ **Reusable clusters:** The amount of time required to create and destroy clusters before and after each test will quickly become overwhelming with reasonable amounts of hardware.
+
+## The Path Forward
+
+### Short Term
+
+#### Test Against Current Branch
+
+The current tests will be modified to run against the latest versions of artifacts. This will be accomplished by removing the existing hardcoded registry links from the existing tests, and implementing a means by which URLs to test-run-specific artifacts will be used. These up-to-date artifacts will be used (at a minimum) when installing the EKS-A Curated Packages controller during cluster creation, as well as when said controller acquires its active bundle. The newly-acquired active bundle will likewise contain links to newly-built test artifacts, such that the end-to-end tests will be run against artifacts built using the current git commits under test, and not the hard-coded development or production registries, as is the case now.
```
junshun, line 41: Will there be a way for a developer to specify their build of package controller to run the local E2E test as an option? Or is the expectation that this will be done in the long term with running e2e test on ok-to-test pull requests?

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+Test runs should be accomplished using development ECR repositories (i.e. public ECR), but if or where that isn't the case, credentials for accessing a private ECR would have to be acquired, which could prove a technical risk, as to my knowledge, this isn't presently done.
```
junshun, line 43: The current E2E test work locally by either exporting the environment variables in the terminal or with a .env file. I believe the E2E test on the release branch use AWS secret manager for other enviroment variables.

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Investigate Daily E2E Run
+
+Running end-to-end tests more frequently is another priority of the short term plan. An investigation into the feasibility of adding a periodic run of the end-to-end tests will be performed. If a periodic end-to-end run is estimated to be within a week's work, then it will be undertaken. The intention is to be notified of end-to-end test failures before a release branch is created, when deadlines are already set and there are possibly many time-sensitive tasks already waiting to be completed.
```
junshun, line 47: I assume these daily E2E runs will be done on the main branch in CP instead of against the latest artifacts right?

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Independent Build Tooling and Testing
+
+In order for new CP end-to-end tests to be as straightforward as possible, and to run with the desired frequency goals, it is necessary to separate our build-tooling processes from those currently existing in the eks-anywhere and eks-anywhere-build-tooling repositories. The current processes produce significant developer friction when writing tests as well as running tests outside of the build pipelines. The existing system's builds are not as finely-grained as is required to meet the test-run-frequency goals defined previously in this document. Lastly, because the concerns of CP tests have little overlap with those of the EKS-A CLI, it is deemed appropriate that these distinctions in build and testing frameworks would allow each project to move more freely in diretions that serves their individual goals best, while any infrastruture or utilities common to both could live in a neutral repository and shared by each.
```
junshun, line 77: Do we need to separate from build-tooling? We could have a seperate process that for example builds main daily so our E2E test runs agains that. We could then continue to use build tooling to cut git tags which specifies that a build is production ready.

docs/testing/e2e-plan.md
ewollesen: In the short term, the plan is to simply run the tests against dev versions via the existing build server. The long term plan includes allowing local runs against arbitrary versions of packages via a custom manifest file.

Depending on how the variables for the build server are passed, it is possible that the ability to specify a local build for local testing would be possible, but I haven't confirmed that, and I'd consider it a bonus, as opposed to a stated goal.

docs/testing/e2e-plan.md
ewollesen: > I assume these daily E2E runs will be done on the main branch in CP instead of against the latest artifacts right?

I was planning for these to run against main. In the context of a daily run, what would be more appropriate? What do you consider "the latest artifacts" if not those built from main?

docs/testing/e2e-plan.md
ewollesen: Long-term, we plan to build far more frequently than just daily, and our builds really shouldn't be tied to or limited by the existing built-tooling tagging system. The tag system is an artifact of how EKS-A is built from a separate repo from the dev repo. I'm sure they have good reasons for that, but by developing and building from the same repo, we simplify our build system. It's easier to trace what's being built, and to ensure that it's as reproducible as possible.

docs/testing/e2e-plan.md
a-cool-train, line 41: I am not sure I follow the mechanisms we obtain the URLS to test run the specific artifacts. The package controller is currently included in the release manifest which is used to fetch the controller to be used. How do we plan on specifying the controller version to use? Would this mean we will be adding curated packages as a dependency to the e2e framework?

a-cool-train, line 41: Also does the short term solution also apply to all the other components currently under test in e2e?

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+Running end-to-end tests more frequently is another priority of the short term plan. An investigation into the feasibility of adding a periodic run of the end-to-end tests will be performed. If a periodic end-to-end run is estimated to be within a week's work, then it will be undertaken. The intention is to be notified of end-to-end test failures before a release branch is created, when deadlines are already set and there are possibly many time-sensitive tasks already waiting to be completed.
+
+### Long Term
+
+Following the short term goals, several broader, more in-depth changes will help our end-to-end tests further raise our confidence. What follows is a brief summary of the larger components or actions required, with a more detailed design after the short term priorities have been satisfied.
+
+#### Test Manifests
+
+A list of each artifact produced, along with identifiers and URLs to acquire each artifact for testing. This manifest will serve as one input to a new testing framework that will be able to ensure that the desired combinations of providers, host OSes, and any other test dimensions deemed appropriate are run.
```
a-cool-train, line 55: Where would the list of the artifacts and the identifiers used be provided? Would there be a mechanism to auto generate this list?

docs/testing/e2e-plan.md
junshun: I do envision a potential scenario where we will be maintaining/patch release branches in which case I imagine we would want to run E2E on in changes in those branches.

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+There will likely be some small number of the current CP E2E tests, particularly those concerned with verifying that a CLI installation of the EKS-A CP controller is successful, that could remain in the eks-anywhere repository. The existing Prow infrastructure will be used, and reporting will be delivered in the same methods to minimize disruption to other systems, habits, or customs previously developed.
```
ivyostosh, line 79: Could you clarify on package specific tests too? Is the expectation to test most of package specific tests without CLI as well?

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+#### Test Configuration Control
+
+A means by which tests can be disabled or enabled based upon their properties, requirements, or current statuses. In plainer words, a way to write a test that doesn't run for a specific provider, or to temporarily disable tests for specific host OS. This input, along with the test manifest are envisioned as being the bulk of the input to a given CP E2E test run.
+
+#### End-To-End Test Documentation
+
+Documentation will be generated that, in addition to the forthcoming design document, will aid developers and other contributors understand the end-to-end test framework, specifically how to write a new end-to-end test, how to run an end-to-end test outside of the build pipelines, how to be notified of failures, and how to debug test failures.
+
+A stretch goal could be to add documentation-style tests in the spirit of [Rust's documentation tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) or [go test examples](https://pkg.go.dev/testing#hdr-Examples).
+
```
ivyostosh, line 74: One other thing I can think of is to enhance test logging and notification logics so that the troubleshooting process for developers become easier.

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
++ **Platform configuration:**
+  A test should be able to exclude specific EKS-A versions, EKS-A providers, cluster node OSes, or K8s versions with which they are incompatible or unsupported. A test will be run on every permutation by default.
+
++ **Configuration review & control:**
+  The test dimensions should be easily viewed, verified, and edited by humans (or reasonable facsimiles thereof). This provides a means of quickly excluding tests categorically for reasons such as hardware outage.
+
+Additionally, it is foreseen that the number of E2E tests that need to be run for each PR will quickly grow, and so it is recommended that these performance goals also be considered:
+
++ **Reusable clusters:** The amount of time required to create and destroy clusters before and after each test will quickly become overwhelming with reasonable amounts of hardware.
```
jonahjon, line 33: Is there any merit to using the remote cluster functionality of Prow here? In that case we could potentially have long running clusters on each provider, and leverage prow for this? https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L41   and can overide the job configuration here. https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-cloud-provider-aws.yaml#L5

docs/testing/e2e-plan.md
junshun: Yes and these type of test should live in eks-anywhere as these are set at cluster creation.

docs/testing/e2e-plan.md
ewollesen: > Is there any merit to using the remote cluster functionality of Prow here? In that case we could potentially have long running clusters on each provider, and leverage prow for this? https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L41 and can overide the job configuration here. https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-cloud-provider-aws.yaml#L5

I had initially had this as something I was hoping we could do, but I was speaking with someone (and geez, I thought it was you) whom said that since our tests were triggered via code builder that they wouldn't be able to run on our prow clusters? I wonder with whom I was having that conversation now?

In any event, prow was one of the first things I wanted to leverage to do all of this. After all, it was built for k8s testing. There could possibly be some issues around having clusters that meet all of the various requirements (bare metal, vs vsphere vs docker etc) all accessible for prow to be able to trigger the jobs, but in general, I like the idea of doing it that way, because it feels like a sort of dog fooding. We build on kubernetes, we run on kubernets, we test on kubernetes, that sort of thing. Plus prow might be a candidate for a package itself one day, and then it would be cool to run tests on a prow instance that was itself setup and run by EKS-A. :D 

There are still questions about the suitability of prow though (at least in my mind), because as Binyam pointed out, we need to be testing first and foremost through eksctl-anywhere. I'm not sure how tests are launched on a prow installation, but if we can't launch our jobs via eksctl-anywhere, then that severely limits its usefulness to us as a test-running platform. I'd be surprised if we couldn't, but it's still an unknown (to me).

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
+# Curated Packages End-To-End Testing Plan
+
+This document summarizes the current state and future plans for end-to-end (E2E) testing in the EKS-A Curated Packages (CP) project.
```
TerryHowe, line 3: every time I see "CP" I think control plane

docs/testing/e2e-plan.md
TerryHowe: We should definitely leave all the hello-app tests over there that exercise all the CLI functionality like generate, list, create cluster, delete cluster, etc.

docs/testing/e2e-plan.md
TerryHowe: Dan spent a fair amount of time creating tools to extract test information. Hopefully, we don't go with a system where we need to develop similar tools

docs/testing/e2e-plan.md
TerryHowe: Examples or is that a follow up?

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+*But the existing tests are not nearly enough to provide confidence that our packages are bug-free.* The current CP E2E framework does not (without significant effort by the test writer) run the existing tests against the versions of their artifacts that matter most[^1]. Nor do they run them for each submitted pull request (PR). These things are not the fault of the tests themselves, nor the tests' writers, but rather they've exposed the limitations of the current CP E2E framework.
+
+[^1]: Early E2E tests were run against production artifacts, generally because they were the only reliably accessible artifacts at the time.
+
+## The End-Goal
```
danbudris, line 15: in these goals, consider including support for debugging test failures -- such as an extension of the existing EKS-A test tool to support the logs/artifacts/etc associate with the new CP dimensions of the system

docs/testing/e2e-plan.md
ewollesen: > every time I see "CP" I think control plane

I've been trying to avoid saying "packages" alone, because it feels so generic, but you make a great point about control plane too. :/ Ugh. Cpackages seems awkward. I dunno.

docs/testing/e2e-plan.md
ewollesen: > Dan spent a fair amount of time creating tools to extract test information. Hopefully, we don't go with a system where we need to develop similar tools

I'm not sure of all the tools you might be referencing, but at least some of the ones I'm aware of are a result of cloudwatch being a little awkward to use... I'm not sure how we avoid that, but your advice is good nonetheless.

docs/testing/e2e-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+To be able to provide confidence that the CP offerings are working across our supported EKS-A versions, EKS-A providers, cluster node OSes, and K8s versions, the current framework must be improved in the following ways:
+
++ **Artifact version control:**
+  Tests should be run against an arbitrary versions of artifacts, defaulting to their latest versions.
+
++ **Timing:**
+  E2E tests should be run on every revision of an "ok to test" PR. This may present some challenges for the current E2E test architecture, as it might require communication between separate AWS pipelines, or failing that, custom code that can communicate with other AWS resources to identify and list available artifacts built in other pipelines.
```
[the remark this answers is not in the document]
jiayiwang7: How expensive is this operation? Do we have some long running eks-a cluster(s) in CI for curated package e2e tests? So the packages are installed and tested on existing clusters.

docs/testing/e2e-plan.md
lewisdiamond: I think "package" is quite fine, even if generic.

TerryHowe (approved):
/lgtm

PR #454, opened 2022-08-31
Generatebundle support long repos

jonahjon:
*Issue #, if available:*

*Description of changes:*

Can support repo names with multiple slashes in the names.    `charts/org/repo`


generatebundlefile/ecr_helper.go
```diff
@@ -73,13 +73,19 @@ func splitECRName(s string) (string, string, error) {
 		if len(chartNameList) == 4 {
 			return fmt.Sprintf("%s/%s", chartNameList[2], chartNameList[3]), chartNameList[3], nil
 		}
```
ivyostosh, line 75: I wonder if ==4 and > 4 scenario can be consolidated to keep the logic clean? I don't see issues with it but let me know if I omit something.

generatebundlefile/ecr_helper.go
```diff
@@ -73,13 +73,19 @@ func splitECRName(s string) (string, string, error) {
...
+		if len(chartNameList) > 4 {
+			return strings.Join(chartNameList[2:], "/"), chartNameList[len(chartNameList)-1], nil
+		}
 	}
 	if len(chartNameList) == 2 {
 		return chartNameList[1], chartNameList[1], nil
 	}
 	if len(chartNameList) == 3 {
```
ivyostosh, line 83: Same comment here for ==3 and >3 scenario. Can they be consolidated?

jonahjon: Yea lemme test that out

jonahjon: ok pushed with c930afcdf3ffeaafdbfc746c124542c2067710b3

Issue #459, opened 2022-09-01
Upgrading packages leads to error `Unsupported value: \"upgrade available\"`


ivyostosh:
**What happened**: Cannot complete a package upgrade

**How to reproduce it (as minimally and precisely as possible)**:
```
oras pull public.ecr.aws/l0g8r8j6/eks-anywhere-packages-bundles:v1-23-latest

k apply -f bundle.yaml
packagebundle.packages.eks.amazonaws.com/v1-23-175 created

eksa upgrade packages --bundle-version v1-23-175
packagebundlecontroller.packages.eks.amazonaws.com/<cluster-name> configured

k logs eks-anywhere-packages-xxx -n eksa-packages
...
1.6620607606287925e+09	ERROR	PackageBundleController	processing bundle controller	{"error": "updating ivyjin-mgmt-5 status to upgrade available: PackageBundleController.packages.eks.amazonaws.com \"ivyjin-mgmt-5\" is invalid: status.state: Unsupported value: \"upgrade available\": supported values: \"ignored\", \"active\", \"disconnected\""}
...

```

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release: built from main on 9/1/2022

ivyostosh:
Fixed by https://github.com/aws/eks-anywhere-packages/pull/460

PR #460, opened 2022-09-01
Add upgrade available

TerryHowe:


ivyostosh:
This should fix https://github.com/aws/eks-anywhere-packages/issues/459. Need to be tested.

PR #467, opened 2022-09-07
Include permanent fix for package webhook

a-cool-train:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/webhook/package_webhook.go
+++ b/pkg/webhook/package_webhook.go
@@ -57,11 +57,6 @@ func (v *packageValidator) Handle(ctx context.Context, request admission.Request
 			fmt.Errorf("decoding request: %w", err))
 	}
 
-	// Temporary fix since testing is currently broken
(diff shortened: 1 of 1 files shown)


pkg/webhook/package_webhook.go
```diff
@@ -127,11 +122,17 @@ func validatePackage(p *v1alpha1.Package, jsonSchema []byte) (*gojsonschema.Resu
 		return nil, fmt.Errorf("error compiling schema %v", err)
 	}
 
-	packageConfig, err := yaml.YAMLToJSON([]byte(p.Spec.Config))
+	packageConfigBytes, err := yaml.YAMLToJSON([]byte(p.Spec.Config))
```
TerryHowe, line 125: Seems a bit odd to parse it, change it to a string and then ignore all that if the thing was blank, but I can see how this would work

TerryHowe (approved):
/lgtm

PR #471, opened 2022-09-09
Bumping resources for controller 50%

jonahjon:

*Issue #, if available:*

*Description of changes:*

50% increase to resources.


lewisdiamond (changes requested):

jonahjon:
/hold
This might be moot since the issue could be from helm looping on create failures.

lewisdiamond:
I'd still push this as there's no real downside if we don't modify the requests.
/lgtm
/approve

vincentni (approved):

PR #475, opened 2022-09-09
Include fix for workload cluster

a-cool-train:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/bundle/client.go
+++ b/pkg/bundle/client.go
@@ -85,7 +85,18 @@ func (bc *bundleClient) GetPackageBundleController(ctx context.Context) (*api.Pa
 	})
 	err := bc.List(ctx, u)
 	if err != nil {
-		return nil, fmt.Errorf("listing KubeadmControlPlane: %v", err)
+		// if an error occurs, it is highly likely this is a workload cluster.
+		// for a workload cluster, fetch the pbcs in the cluster
+		// since there will only be a single pbc currently, we use the first pbc found
+		pbcList := &api.PackageBundleControllerList{}
+		err = bc.List(ctx, pbcList)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/bundle/client.go
```diff
@@ -85,7 +85,18 @@ func (bc *bundleClient) GetPackageBundleController(ctx context.Context) (*api.Pa
 	})
 	err := bc.List(ctx, u)
 	if err != nil {
```
lewisdiamond, line 87: Just log this error as a warning and remove the rest of the code. Line 103 `name := api.PackageBundleControllerName` will default it.

a-cool-train: But this will not require the user to modify their pbcs, that's why I added the listing of the pbcs that currently exist in the cluster. Currently, there will only be 1.

pkg/bundle/client.go
```diff
@@ -85,7 +85,18 @@ func (bc *bundleClient) GetPackageBundleController(ctx context.Context) (*api.Pa
...
+		// since there will only be a single pbc currently, we use the first pbc found
+		pbcList := &api.PackageBundleControllerList{}
+		err = bc.List(ctx, pbcList)
+		if err != nil {
+			return nil, fmt.Errorf("listing Packagecontrollers: %v", err)
+		}
+		items := pbcList.Items
+		if len(items) > 0 {
+			return &items[0], nil
+		}
```
TerryHowe, line 99:
```suggestion
		}
		return nil, fmt.Errorf("listing KubeadmControlPlane: %v", err)
```

TerryHowe: Something of that sort if there is no PBC?

TerryHowe (approved):
/lgtm
/approve

PR #482, opened 2022-09-20
File handling for kubeconfig

TerryHowe:
I'd eventually like to get away from creating the files, but the way helm works this is a lot easier.

### DIFF
--- a/pkg/authenticator/kubeconfig.go
+++ b/pkg/authenticator/kubeconfig.go
@@ -43,6 +43,11 @@ func (kc *kubeconfigClient) GetKubeconfig(ctx context.Context, clusterName strin
 		return "", fmt.Errorf("getting kubeconfig for cluster %q: %w", clusterName, err)
 	}
 
+	if _, err := os.Stat(secretName); err == nil {
+		_ = os.Chown(secretName, os.Geteuid(), os.Getegid())
+		_ = os.Chmod(secretName, 0600)
+
+	}
 	err = os.WriteFile(secretName, kubeconfigSecret.Data["value"], 0600)
 	if err != nil {
 		return "", fmt.Errorf("opening temporary file: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/authenticator/kubeconfig_test.go
```diff
@@ -35,6 +35,7 @@ func TestManagerContext_getRegistry(t *testing.T) {
 		fileName, err := sut.GetKubeconfig(ctx, "billy")
 		assert.NoError(t, err)
 		assert.Equal(t, "billy-kubeconfig", fileName)
-		os.Remove(fileName)
+		err = os.Remove(fileName)
+		assert.NoError(t, err)
```
ivyostosh, line 39: Should we use `require.NoError` more consistently across different tests?

TerryHowe: This is just informative on the clean-up. The require.NoError stops the tests which would be fine here, but this error is not very important

pkg/authenticator/kubeconfig_test.go
```diff
@@ -35,6 +35,7 @@ func TestManagerContext_getRegistry(t *testing.T) {
 		fileName, err := sut.GetKubeconfig(ctx, "billy")
 		assert.NoError(t, err)
```
ewollesen, line 36: (This one could probably be a require.NoError though 😁)

pkg/authenticator/kubeconfig.go
```diff
@@ -43,6 +43,11 @@ func (kc *kubeconfigClient) GetKubeconfig(ctx context.Context, clusterName strin
 		return "", fmt.Errorf("getting kubeconfig for cluster %q: %w", clusterName, err)
 	}
 
+	if _, err := os.Stat(secretName); err == nil {
+		_ = os.Chown(secretName, os.Geteuid(), os.Getegid())
```
ewollesen, line 47: I would have returned these as errors. Better to know sooner than later that there's a problem.

I'm guessing you disagree, but would you be amenable to at least a log message, so we'd have some clue that there's something wrong?

TerryHowe: 99% sure the error would be permission denied which will show up below on the write. I don't want to clutter this up with code I think is a bit throw away away. Hopefully, we move away from this file approach.

PR #488, opened 2022-09-24
Move sort into bundle list

TerryHowe:
The get bundle list method needs to be a bit more selective for remote management (by k8s version), so this takes it one baby step by combining the get with the sort.


vincentni (approved):
/lgtm
/approve

PR #492, opened 2022-09-26
Move active bundle download

TerryHowe:
Move the active bundle download to PBC reconciliation

The reason for the move is if we were to reconcile on a deleted bundle, we'd have to iterate through all the PBCs to find out if it was an active bundle. That is possible, but I'd sooner the dependency went one way PBC knows about bundles, but bundles don't know necessarily about PBCs.

### DIFF
--- a/pkg/bundle/client.go
+++ b/pkg/bundle/client.go
@@ -24,6 +24,9 @@ type Client interface {
 	// GetBundleList get list of bundles worthy of consideration
 	GetBundleList(ctx context.Context, serverVersion string) (bundles []api.PackageBundle, err error)
 
+	// GetBundle retrieves the named bundle.
+	GetBundle(ctx context.Context, name string) (namedBundle *api.PackageBundle, err error)
+
 	// CreateBundle add a new bundle custom resource
 	CreateBundle(ctx context.Context, bundle *api.PackageBundle) error
 
@@ -87,6 +90,27 @@ func (bc *bundleClient) GetPackageBundleController(ctx context.Context, clusterN
 		return nil, fmt.Errorf("getting PackageBundleController: %v", err)
 	}
 	return &pbc, nil
+}
+
+func (bc *bundleClient) GetBundle(ctx context.Context, name string) (namedBundle *api.PackageBundle, err error) {
+	namedBundle = &api.PackageBundle{}
+	if name == "" {
+		return namedBundle, nil
+	}
+
+	nn := types.NamespacedName{
+		Namespace: api.PackageNamespace,
+		Name:      name,
+	}
+	err = bc.Get(ctx, nn, namedBundle)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	return namedBundle, nil
 }
 
 func (bc *bundleClient) GetBundleList(ctx context.Context, serverVersion string) (bundles []api.PackageBundle, err error) {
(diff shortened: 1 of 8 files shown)


pkg/bundle/manager.go
```diff
@@ -109,6 +109,28 @@ func (m *bundleManager) ProcessBundleController(ctx context.Context, pbc *api.Pa
 			return fmt.Errorf("creating namespace for %s: %s", pbc.Name, err)
 		}
 
+		activeBundle, err := m.bundleClient.GetBundle(ctx, pbc.Spec.ActiveBundle)
+		if err != nil {
+			m.log.Error(err, "Unable to get active bundle", "bundle", pbc.Spec.ActiveBundle)
+			return nil
+		}
+
+		if activeBundle == nil {
```
ewollesen, line 118: You don't handle the case where activeBundle isn't nil, but also isn't set, as would be returned from GetBundle(ctx, "") in pkg/bundle/client.go:148. You return an initialized value there, so it won't match as nil.

Also, I don't see a test below for that scenario, though there might be an existing test that I can't see cuz github.

pkg/bundle/manager_test.go
```diff
@@ -128,6 +128,60 @@ func TestBundleManager_ProcessBundleController(t *testing.T) {
...
 		bc.EXPECT().CreateClusterNamespace(ctx, pbc.Name).Return(nil)
+		bc.EXPECT().GetBundle(ctx, pbc.Spec.ActiveBundle).Return(&allBundles[0], nil)
+
+		err := bm.ProcessBundleController(ctx, pbc)
+
+		assert.NoError(t, err)
+		assert.Equal(t, api.BundleControllerStateActive, pbc.Status.State)
+	})
+
+	t.Run("active missing active bundle", func(t *testing.T) {
```
ewollesen, line 139: As mentioned above, I don't see a test that exercises the path when the active bundle is "".

ewollesen:
I'd appreciate more context in the commit message. You're moving the active bundle download, ok, but why? What do we gain? Why is this an improvement or necessary?

TerryHowe:
> I'd appreciate more context in the commit message. You're moving the active bundle download, ok, but why? What do we gain? Why is this an improvement or necessary?


The reason for the move is if we were to reconcile on a deleted bundle, we'd have to iterate through all the PBCs to find out if it was an active bundle. That is possible, but I'd sooner the dependency went one way PBC knows about bundles, but bundles don't know necessarily about PBCs.

pkg/bundle/manager.go
TerryHowe: Changed this to something more logical

pkg/bundle/client.go
```diff
@@ -140,6 +143,27 @@ func (bc *bundleClient) GetActiveBundleNamespacedName(ctx context.Context) (type
 	return nn, nil
 }
 
+func (bc *bundleClient) GetBundle(ctx context.Context, name string) (namedBundle *api.PackageBundle, err error) {
+	namedBundle = &api.PackageBundle{}
+	if name == "" {
+		return namedBundle, nil
```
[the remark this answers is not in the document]
TerryHowe: Got rid of this.

pkg/bundle/manager_test.go
TerryHowe: moved the "" activeBundle to manager

TerryHowe: These tests aren't for activeBundle=="", but for just a missing active bundle

pkg/bundle/manager.go
```diff
@@ -81,7 +81,7 @@ func (m *bundleManager) ProcessBundleController(ctx context.Context, pbc *api.Pa
 		return nil
 	}
 
-	latestBundle, err := m.registryClient.LatestBundle(ctx, pbc.GetBundleUri(), info.String())
+	latestBundle, err := m.registryClient.LatestBundle(ctx, pbc.GetBundleURI(), info.String())
```
jonahjon, line 84: Don't actually see any change here?

TerryHowe: URI toUpper

Issue #433, opened 2022-08-19
webhook message not returned to user in feedback


TerryHowe:
We have useful webhook messages like `activeBundle <%q> not present on cluster` but these are not returned to the user.

TerryHowe:
Fixed

PR #493, opened 2022-09-27
Use status reason for description

TerryHowe:
It looks like the k8s community had some confusion over reason and message. The comments indicate message is supposed to be machine readable and reason is supposed to be human readable, but the usage is opposite that. For the customer, reason is displayed, so for example before this change:
```
% k apply -f pbctlhowe.yaml 
The request is invalid
```
After this change:
```
% k apply -f pbctlhowe.yaml
Error from server (activeBundle "" not present on cluster): error when creating "pbctlhowe.yaml": admission webhook "vpackagebundlecontroller.kb.io" denied the request: activeBundle "" not present on cluster
```


TerryHowe:
Closes: https://github.com/aws/eks-anywhere-packages/issues/433

pkg/webhook/packagebundlecontroller_webhook.go
```diff
@@ -92,15 +101,15 @@ func (v *activeBundleValidator) handleInner(_ context.Context, pbc *v1alpha1.Pac
 	}
 
 	if !found {
-		msg := fmt.Sprintf("activeBundle <%q> not present on cluster", pbc.Spec.ActiveBundle)
+		reason := fmt.Sprintf("activeBundle %q not present on cluster", pbc.Spec.ActiveBundle)
```
ewollesen, line 104:
```suggestion
		reason := fmt.Sprintf("active bundle %q is not present on cluster", pbc.Spec.ActiveBundle)
```
I like dropping the <>, and since this is for a human, I'd take it this little step further.

Would it make sense/be simple to add:
```suggestion
		reason := fmt.Sprintf("active bundle %q is not present on cluster %q", pbc.Spec.ActiveBundle, clusterName)
```
?? I don't know if the clusterName is handy in the environment.

ewollesen (commented):
Just one small suggestion, take it or leave it

/lgtm
(no slash approve so you can merge as is, or make more changes, your choice)

pkg/webhook/packagebundlecontroller_webhook_test.go
```diff
@@ -159,7 +159,7 @@ func TestHandleInner(t *testing.T) {
 		if assert.NoError(t, err) {
 			assert.False(t, resp.AdmissionResponse.Allowed)
 			assert.Equal(t, metav1.StatusFailure, resp.AdmissionResponse.Result.Status)
-			assert.Equal(t, metav1.StatusReasonInvalid, resp.AdmissionResponse.Result.Reason)
+			assert.Equal(t, "activeBundle \"v1-21-1002\" not present on cluster", string(resp.AdmissionResponse.Result.Reason))
```
TerryHowe, line 162:
```suggestion
			assert.Equal(t, "active bundle \"v1-21-1002\" not present on cluster", string(resp.AdmissionResponse.Result.Reason))
```

TerryHowe:
had to rebase again

PR #495, opened 2022-09-28
Update to go.18

TerryHowe:


scripts/common.sh
```diff
@@ -198,6 +198,10 @@ function build::common::get_go_path() {
   if [[ $version == "1.17"* ]]; then
     gobinaryversion="1.17"
   fi
+  if [[ $version == "1.18"* ]]; then
```
ewollesen, line 201:
```suggestion
  if [[ "$version" == "1.18"* ]]; then
```
The double quotes protect you in case $version is empty. You'll get a sometimes difficult to debug error message from bash.

PR #501, opened 2022-09-30
Updating ecrtokenrefresher to support updating secret on workload clusters

junshun:
Updating ecr-token-refresher to look for workload clusters via parsing namespaces. Token push was also updated to target each of the workload clusters using a kubeconfig via a secret.


TerryHowe (approved):

TerryHowe (commented):
/lgtm

PR #503, opened 2022-09-30
Handle get PBC and active bundle

TerryHowe:
Get a PBC for a specific cluster

* Pass clusterName to the get PBC method
* Do not use kubeadmcontrolplane
* Get active bundle based on PBC for the cluster
* Do not reconcile activeBundle delete on bundle reconciliation

### DIFF
--- a/api/v1alpha1/package.go
+++ b/api/v1alpha1/package.go
@@ -1,12 +1,18 @@
 package v1alpha1
 
 import (
+	"os"
+	"strings"
+
 	"sigs.k8s.io/yaml"
 )
 
 const (
-	PackageKind      = "Package"
-	PackageNamespace = "eksa-packages"
+	PackageKind       = "Package"
+	PackageNamespace  = "eksa-packages"
+	namespacePrefix   = PackageNamespace + "-"
+	oldPbcName        = "bundle-controller"
+	clusterNameEnvVar = "CLUSTER_NAME"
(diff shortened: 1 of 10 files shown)


api/v1alpha1/package.go
```diff
@@ -23,3 +29,24 @@ func (config *Package) GetValues() (values map[string]interface{}, err error) {
 	err = yaml.Unmarshal([]byte(config.Spec.Config), &mapInterfaces)
 	return mapInterfaces, err
 }
+
+func (config *Package) GetClusterName() string {
+	if strings.HasPrefix(config.Namespace, namespacePrefix) {
+		clusterName := strings.TrimPrefix(config.Namespace, namespacePrefix)
+		// Backward compatibility
+		if clusterName == oldPbcName {
+			return os.Getenv(clusterNameEnvVar)
```
jonahjon, line 38: How exactly does this work?

jonahjon:
Not much code changed just re-arranged. Had one Q but otherwise it looked good.

api/v1alpha1/package.go
TerryHowe: If the installation has the old PBC name which is something like `eksa-package-bundle-controller`, then use the environment variable to get the cluster name. I probably will follow up with a PR to remove support for this and some migration.

PR #509, opened 2022-10-04
Add Adot sample config

ivyostosh:

### DIFF
--- a/sample_package_configs/adot/daemonset_default/daemonset_default.yaml
+++ b/sample_package_configs/adot/daemonset_default/daemonset_default.yaml
@@ -1,3 +1,8 @@
+#
+# ADOT Package configuration
+# The default package configuration deploys an ADOT Collector with the metrics 
+# pipeline, which includes otlp and prometheus receiver, and logging exporter.
+#
 apiVersion: packages.eks.amazonaws.com/v1alpha1
 kind: Package
 metadata:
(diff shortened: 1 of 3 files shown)


sample_package_configs/adot/daemonset_default/daemonset_default.yaml
```diff
@@ -0,0 +1,15 @@
+#
```
ivyostosh, line 1: The reason why I'm creating separate folders here is because I want to add rendered results later in each folder as well.

jonahjon:
Why does this need to live in this repo? Shouldn't this belong in docs

jonahjon:
We don't have much of anything in here that is package specific.

ivyostosh:
The issue I have with docs is that it's really hard for customers to customize what they want to use just through copy and paste what we have offered there. The most valuable piece ADOT offers is the ability for customers to customize any receiver, collector, and exporter they want. And we should offer them various examples that are easy for them to leverage. This may be different from other packages we have today.

TerryHowe:
Well adot doesn't have to follow the form in docs like everything else. Perhaps this is more of a workshop thing? There is only a Harbor workshop, but maybe an adot? I know people think the docs is too spread out, so that wouldn't help with that though.  Other than that, add it to tasks or reference, but no one is going to find it here

/hold

ivyostosh:
All the configs here does exit in our doc site, which includes a workshop, so ADOT is following all conventions for docs similar to any other packages. A question was raised during workshop review about how can we make it easy for customers to modify configs (and again, ADOT is all about configs). A long-term solution is to modify generate packages command to support multiple pre-defined configs, but we are not there yet. So we agreed on a short-term solution to park configs here for customers to grab (and I will reference here from the doc). I'm happy to consider other alternatives if you have other suggestions.

TerryHowe:
Perhaps download feature of Hugo would help for this and it keeps the docs all in one repo

PR #511, opened 2022-10-05
Package Controller update to refresh tokens for Remote Management

junshun:
The following changes were done
**Create configmap in target namespace for workload cluster** - this configmap is used by the ecr-token-refresher job to update secrets to specific namespaces
**Change target namespace for adding/deleting from config map** - this is needed to determine which cluster to add/remove from
**Changed ecr secret to trigger job instead of manually adding secrets to namespaces** - This is done to simplify the process and have the job handle it all. With this we will also trigger the job twice. Once before package installed and once after. The after case is to catch if a namespace is created by the helm chart itself.


junshun:
/hold
Still need to fix some of the bundle client test

junshun:
Closing as other smaller PRS cover these changes

PR #513, opened 2022-10-06
Create Configmap in Manager for Remote Management

junshun:
Portion of remote management changes to support the ecr-token-refresher. This change is to create the config maps in the corresponding namespace eksa-packages-${CLUSTER_NAME} so token refresher can push to.
Example if management cluster is named mgmt, we will be creating a configmap in eksa-packages-mgmt which holds the namespaces we need to push ecr-tokens to. Config map would look like this

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: ns-secret-name
  namespace: eksa-packages-mgmt
data:
  eksa-packages-mgmt: eksa-packages-placeholder
  emissary-system: emissary-ingress
```

### DIFF
--- a/pkg/bundle/client.go
+++ b/pkg/bundle/client.go
@@ -8,10 +8,12 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+	auth "github.com/aws/eks-anywhere-packages/pkg/authenticator"
 )
 
 type Client interface {
@@ -32,6 +34,9 @@ type Client interface {
 
 	// CreateClusterNamespace based on cluster name
 	CreateClusterNamespace(ctx context.Context, clusterName string) error
+
+	// CreateClusterConfigMap based on cluster name
+	CreateClusterConfigMap(ctx context.Context, clusterName string) error
 
 	// SaveStatus saves a resource status
 	SaveStatus(ctx context.Context, object client.Object) error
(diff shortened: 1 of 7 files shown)


pkg/bundle/client.go
```diff
@@ -155,6 +160,41 @@ func (bc *bundleClient) CreateClusterNamespace(ctx context.Context, clusterName
...
+	}
+
+	if !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	cm.Data = make(map[string]string)
+	cm.Data[namespace] = "eksa-package-controller"
+	// Unfortunate workaround for emissary webhooks hard coded crd namespace
+	cm.Data["emissary-system"] = "eksa-package-placeholder"
```
ivyostosh, line 189: Did we do a check on existing packages and packages that will go out in this release, and confirm emissary is the only one that needs this workaround?

junshun: New Packages so far do not create any namespaces. Exception was cert-manager however there is a patch for that.

a-cool-train, line 189: Any reason why the namespace is eksa-package-placeholder?

pkg/bundle/client.go
```diff
@@ -155,6 +160,41 @@ func (bc *bundleClient) CreateClusterNamespace(ctx context.Context, clusterName
...
+	}
+
+	cm := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+	err := bc.Get(ctx, key, cm)
+	if err == nil {
```
a-cool-train, line 178: Also in go, we always return from an error scenario and let the program flow unless an error is encountered.

pkg/bundle/client.go
```diff
@@ -155,6 +160,41 @@ func (bc *bundleClient) CreateClusterNamespace(ctx context.Context, clusterName
...
+		return nil
```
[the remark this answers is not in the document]
junshun: This portion is similar to the function CreateClusterNamespace already in the file. So whats going on here is we are checking if the Namespace/Config Map already exists. If it does we dont need to do anything when reconciling here. Thus a success and no error. So going into the next line of the code we check if the error is type IsNotFound. If it is thats fine and we can create. Otherwise its an actual error and we need to exit the function with error.

pkg/bundle/client.go
junshun: See previous comment explaining

pkg/bundle/client.go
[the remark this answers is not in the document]
a-cool-train: Can you add that as a comment in the code?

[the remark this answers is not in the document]
junshun: Added

TerryHowe (approved):
/lgtm

PR #519, opened 2022-10-07
Adding promote option from file for cluster autoscaler

jonahjon:

*Issue #, if available:*

*Description of changes:*


generatebundlefile/data/autoscaler/input_promote.yaml
```diff
@@ -0,0 +1,32 @@
+# This file is only used to determine which Helm charts to move from Dev to Prod during a release.
+name: "v1-21-1001"
+kubernetesVersion: "1.21"
```
jonathanmeier5, line 3: Based on offline convo with @jonahjon this `kubernetesVersion` is ignored when this file is used in the promotion process.

This is not a strong opinion and it shouldn't block a merge, but I think it could be confusing to someone jumping into the project. Would we be open to removing this and the bundle name field, provided it doesn't cause problems elsewhere?

PR #520, opened 2022-10-10
Add workloadonly flag

a-cool-train:
*Description of changes:*
- This is part of cert manager package installation. Cert manager will only be installed on a workload cluster and hence this flag is necessary.

### DIFF
--- a/api/v1alpha1/packagebundle_types.go
+++ b/api/v1alpha1/packagebundle_types.go
@@ -49,6 +49,10 @@ type BundlePackage struct {
 	// +kubebuilder:validation:Required
 	// Source location for the package (probably a helm chart).
 	Source BundlePackageSource `json:"source"`
+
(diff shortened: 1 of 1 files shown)


TerryHowe:
This all looks fine for me, I just wonder if there are any other approaches that would make things easier. I'm not convinced there are, but want to discuss. Array of enums for example in a flags field:
```
flags:
  - workdloadOnly
```

lewisdiamond:
> This all looks fine for me, I just wonder if there are any other approaches that would make things easier. I'm not convinced there are, but want to discuss. Array of enums for example in a flags field:
> 
> ```
> flags:
>   - workdloadOnly
> ```

The array of enums would still require a CRD change right? I suppose we could also detach ourselves from CRDs a bit for things like that by adding a "config" field which could be just a string we parse.

PR #521, opened 2022-10-10
Trigger Job when installing packages before and after install.

junshun:
To support updating ecr-token on remote clusters an instance of the cronjob is triggered. 

Two instances of the job are called. The first is to get the ecr-token into the namespaces before an install occurs so that the pods do not back off. A second instance is called to cover namespaces that are created via the helm charts.

Removed nsReleaseMap from ecrsecret as its no longer being used as the job handles keeping track of namespaces. Changed to accept config over rest client as corev1 path are not compatible with batchv1.


TerryHowe (approved):
/lgtm

PR #522, opened 2022-10-10
Fix update available problem

TerryHowe:
Eric made a comment about passing the server version to the get bundle list and I agreed that it was problematic. This PR covers that concern and it fixes an issue where a PBC gets marked "upgrade available" incorrectly.


pkg/bundle/manager.go
```diff
@@ -66,6 +66,15 @@ func (m bundleManager) ProcessBundle(ctx context.Context, newBundle *api.Package
 	return false, nil
 }
 
+func (m *bundleManager) hasLatestBundle(sortedBundles []api.PackageBundle, latestBundle *api.PackageBundle) bool {
```
ewollesen, line 69: This method isn't really checking if sortedBundles has the latest bundle. It's checking that sortedBundles has whichever bundle I pass in the latestBundle argument.

I'd suggest either renaming it to just `hasBundle`, or removing the `latestBundle` argument, and determining what the latest bundle is within the method itself, so as to enforce the "latest" in the name.

Also, I [opened a PR](https://github.com/TerryHowe/eks-anywhere-packages/pull/3) that tweaks some of the assumptions around sorting. I think it's better to do the sorting here if that's what's needed, rather than assuming that GetBundleList will slice (since there's no guarantee or checking that assumption will hold).

pkg/bundle/manager.go
```diff
@@ -66,6 +67,16 @@ func (m bundleManager) ProcessBundle(ctx context.Context, newBundle *api.Package
 	return false, nil
 }
 
+func (m *bundleManager) hasBundleNamed(bundles []api.PackageBundle, bundleName string) bool {
+	sort.Sort(api.BundlesByVersion(bundles))
```
ewollesen, line 71: Now that I see it all in context... Is it even necessary to sort these?

ewollesen (approved):
/lgtm

PR #523, opened 2022-10-10
Active bundles

TerryHowe:
Do not get the active bundle since we just got all of them in a slice

### DIFF
--- a/pkg/bundle/client.go
+++ b/pkg/bundle/client.go
@@ -65,7 +65,7 @@ func (bc *bundleClient) GetActiveBundle(ctx context.Context, clusterName string)
 		return nil, err
 	}
 
-	if pbc.Spec.ActiveBundle == "" {
+	if len(pbc.Spec.ActiveBundle) < 1 {
 		return nil, fmt.Errorf("no activeBundle set in PackageBundleController")
 	}
 
--- a/pkg/bundle/client.go
+++ b/pkg/bundle/client.go
@@ -65,7 +65,7 @@ func (bc *bundleClient) GetActiveBundle(ctx context.Context, clusterName string)
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/bundle/client.go
```diff
@@ -65,7 +65,7 @@ func (bc *bundleClient) GetActiveBundle(ctx context.Context, clusterName string)
 		return nil, err
 	}
 
-	if pbc.Spec.ActiveBundle == "" {
+	if len(pbc.Spec.ActiveBundle) < 1 {
```
ivyostosh, line 68:
```suggestion
	if len(pbc.Spec.ActiveBundle) == 0 {
```

This seems to be a more consistent write-up with the `pkg.bundle/manager.go` file, and easier to understand IMO since you don't really expect the length of bundle to ever go below 0 right?

pkg/webhook/packagebundlecontroller_webhook.go
```diff
@@ -77,7 +77,7 @@ func (v *activeBundleValidator) Handle(ctx context.Context, req admission.Reques
 func (v *activeBundleValidator) handleInner(ctx context.Context, pbc *v1alpha1.PackageBundleController, bundles *v1alpha1.PackageBundleList) (
 	*admission.Response, error) {
 
-	if pbc.Spec.ActiveBundle == "" {
+	if len(pbc.Spec.ActiveBundle) < 1 {
```
ivyostosh, line 80:
```suggestion
	if len(pbc.Spec.ActiveBundle) == 0 {
```

Same comment here

pkg/bundle/client.go
ivyostosh: I agree with what Eric says here. I was thinking there must be some weird behavior on the comparator that triggered this change. If there are no such constraint, I would vote for `""` as well.

PR #525, opened 2022-10-11
Deprecate the old package namespace

TerryHowe:


pkg/packages/manager_test.go
```diff
@@ -181,6 +181,18 @@ func TestManagerLifecycle(t *testing.T) {
 		thenManagerContext(t, mc, api.StateInstalled, expectedSource, 60*time.Second, "")
 	})
 
+	t.Run("installing installs in deprecated namespace", func(t *testing.T) {
```
ivyostosh, line 184: What does installing installs mean?

TerryHowe:
```suggestion
	t.Run("installing in deprecated namespace", func(t *testing.T) {
```

PR #529, opened 2022-10-11
Add ADOT to prod bundle

ivyostosh:

### DIFF
--- a/generatebundlefile/data/input_120_prod.yaml
+++ b/generatebundlefile/data/input_120_prod.yaml
@@ -9,6 +9,13 @@ packages:
         registry: public.ecr.aws/eks-anywhere
         versions:
             - name: 0.1.2-latest-helm
+  - org: aws-observability
+    projects:
+      - name: adot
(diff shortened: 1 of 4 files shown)


generatebundlefile/data/input_120_prod.yaml
```diff
@@ -9,6 +9,13 @@ packages:
         registry: public.ecr.aws/eks-anywhere
         versions:
             - name: 0.1.2-latest-helm
+  - org: aws-observability
+    projects:
+      - name: adot
+        repository: adot/charts/aws-otel-collector
+        registry: public.ecr.aws/eks-anywhere
+        versions:
+            - name: 0.21.1-971c31238aeea5f1bda6ec95caaf24993b304157
```
a-cool-train, line 18: Is this required to be hard coded? Would it be possible to just use latest here?

ivyostosh: For prod, it was intentional to make it hard coded so we don't auto bundle a version.

PR #534, opened 2022-10-11
Add support for package dependencies

lewisdiamond:
*Issue #, if available:*

*Description of changes:*
Adds a dependencies list (string) in `Bundle.Spec.Packages[].Source.Version[]` which refer to package names that must be installed before installing that package. The dependencies are automatically installed with no configuration, in the same namespace as the original package (the namespace of the Package resource). ~No target namespace is defined for dependencies, so it is only useful for cluster-wide resources (for now).~


api/v1alpha1/packagebundle.go
```diff
@@ -26,22 +26,58 @@ func (config *PackageBundle) ExpectedKind() string {
...
-				// versions, the order in the file will be ordered according to what we want `latest` to point to
-				if packageVersion.Name == pkgVersion || packageVersion.Digest == pkgVersion || pkgVersion == Latest {
-					retSource = PackageOCISource{Registry: source.Registry, Repository: source.Repository, Digest: packageVersion.Digest, Version: packageVersion.Name}
-					return retSource, nil
-				}
-			}
+			return pkg, nil
+		}
+	}
+	return retPkg, fmt.Errorf("package not found in bundle (%s): %s", config.ObjectMeta.Name, pkgName)
```
a-cool-train, line 35: I don't see a definition for retPkg. Am I missing something here? And should we be returning nil instead here?

api/v1alpha1/packagebundle.go
```diff
@@ -26,22 +26,58 @@ func (config *PackageBundle) ExpectedKind() string {
...
+		// We do not sort before getting `latest` because there will be only a single packageVersion per release in normal cases. For edge cases which may require multiple
+		// versions, the order in the file will be ordered according to what we want `latest` to point to
+		if packageVersion.Name == pkgVersion || packageVersion.Digest == pkgVersion || pkgVersion == Latest {
+			return packageVersion, nil
 		}
 	}
+	return ret, fmt.Errorf("package version not found in bundle (%s): %s @ %s", config.ObjectMeta.Name, pkg.Name, pkgVersion)
+}
+
+func (config *PackageBundle) FindOCISourceByName(pkgName string, pkgVersion string) (retSource PackageOCISource, err error) {
```
a-cool-train, line 62: Would it make sense to document all exported methods? Specifically the expected behavior. This could improve readability immensely.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: public.ecr.aws/j0a1m4z9/eks-anywhere-packages
+  newName: controller
```
a-cool-train, line 15: What is this change for? Can we add a comment next to it?

pkg/packages/manager.go
```diff
@@ -65,11 +71,67 @@ func (mc *ManagerContext) getRegistry(values map[string]interface{}) string {
...
+			if items[i].Spec.PackageName == dep.Name {
+				pkg = &items[i]
+			}
+		}
+		if pkg != nil {
+			if pkg.Status.State != api.StateInstalled {
+				pkgsNotReady = append(pkgsNotReady, *pkg)
+			}
+		} else {
+			p := api.NewPackage(dep.Name, dep.Name, mc.Package.ObjectMeta.Namespace)
```
a-cool-train, line 116: No need to reference ObjectMeta. `mc.Package.Namespace` should work.

config/manager/kustomization.yaml
lewisdiamond: This is Terry's personal public registry.

api/v1alpha1/packagebundle.go
lewisdiamond: Named returns. See the method definition.

pkg/utils/func.go
```diff
@@ -0,0 +1,9 @@
+package utils
```
TerryHowe, line 1: I generally don't see much value in utils like this. It just doesn't get reused.

pkg/packages/manager.go
```diff
@@ -65,11 +71,67 @@ func (mc *ManagerContext) getRegistry(values map[string]interface{}) string {
...
+			err := mc.PackageClient.CreatePackage(mc.Ctx, &p)
+			if err != nil {
+				mc.Log.Error(err, "creating dependency package")
+			}
+		}
+	}
+
+	if len(pkgsNotReady) > 0 {
+		depsStr := utils.Map(pkgsNotReady, func(pkg api.Package) string { return pkg.Spec.PackageName })
+		mc.Package.Status.Detail = "Waiting for dependencies: " + strings.Join(depsStr, ", ")
```
TerryHowe, line 127: cool

pkg/packages/manager.go
```diff
@@ -65,11 +71,67 @@ func (mc *ManagerContext) getRegistry(values map[string]interface{}) string {
...
+func processInstallingDependencies(mc *ManagerContext) bool {
+	mc.Log.Info("Installing dependencies", "chart", mc.Source)
+	dependencies, err := mc.Bundle.GetDependencies(mc.Version)
+	if err != nil {
+		mc.Package.Status.Detail = fmt.Sprintf("invalid package bundle. (%s@%s bundle: %s)", mc.Package.Name, mc.Version, mc.Bundle.Name)
+		mc.Log.Info(mc.Package.Status.Detail)
+		mc.RequeueAfter = retryLong
+		return true
+	}
+	pkgs, err := mc.PackageClient.GetPackageList(mc.Ctx, mc.Package.Namespace)
```
TerryHowe, line 95: some more descriptive name would help here like installedPackages

api/v1alpha1/packagebundle.go
```diff
@@ -26,22 +26,58 @@ func (config *PackageBundle) ExpectedKind() string {
...
-				}
-			}
+			return pkg, nil
+		}
+	}
+	return retPkg, fmt.Errorf("package not found in bundle (%s): %s", config.Name, pkgName)
+}
+
+func (config *PackageBundle) GetDependencies(version SourceVersion) (dependencies []BundlePackage, err error) {
+	var ret []BundlePackage
```
TerryHowe, line 39: I think you could use dependencies here that you declared as a return

api/v1alpha1/packagebundle.go
```diff
@@ -26,22 +26,58 @@ func (config *PackageBundle) ExpectedKind() string {
...
+	for _, packageVersion := range source.Versions {
+		// We do not sort before getting `latest` because there will be only a single packageVersion per release in normal cases. For edge cases which may require multiple
+		// versions, the order in the file will be ordered according to what we want `latest` to point to
+		if packageVersion.Name == pkgVersion || packageVersion.Digest == pkgVersion || pkgVersion == Latest {
+			return packageVersion, nil
 		}
 	}
+	return ret, fmt.Errorf("package version not found in bundle (%s): %s @ %s", config.Name, pkg.Name, pkgVersion)
+}
+
```
TerryHowe, line 61: Kind of unrelated to this PR, but I was feeling that OCISource was not enough information for the state machine.

pkg/utils/func.go
lewisdiamond: I'd usually agree but when something like `map` isn't in the language, this seems like a necessity.

pkg/packages/manager.go
lewisdiamond: for `GetPackageList` or `pkg`? We don't filter or know about the state at this point.

api/v1alpha1/packagebundle.go
lewisdiamond: Yep that's right.

lewisdiamond:
/hold
Testing fixes related to schema validation

pkg/packages/manager.go
TerryHowe: for pkg

TerryHowe (approved):
/lgtm

PR #535, opened 2022-10-11
Remove extra go vet calls

ewollesen:
The go vet tool, instead of being run explicitly, is merged into the test target. Since go test already runs a chunk of the same checks as go vet[^1], it's more efficient to run it when testing, which is done via the addition of the -vet=all flag to the go test invocation.

Additionally, the go mod tidy call that was part of the vet target isn't required or really useful there, so it's broken out, using Make's built in dependency tracking to trigger it only when needed, during the build target.

[^1]: https://pkg.go.dev/cmd/go@master#hdr-Test_packages

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ else
 GOBIN=$(shell $(GO) env GOBIN)
 endif
 
-all: vet generate manifests build helm/package test # lint is run via github action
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

PR #536, opened 2022-10-12
Feature cert manager changes

a-cool-train:
*Description of changes:*
If a package is intended to only be installed on a workload cluster, deny request if a custom resource is created with intention of installing in management cluster.

### DIFF
--- a/pkg/webhook/package_webhook.go
+++ b/pkg/webhook/package_webhook.go
@@ -63,7 +63,7 @@ func (v *packageValidator) Handle(ctx context.Context, request admission.Request
 		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("getting PackageBundle: %v", err))
 	}
 
-	isConfigValid, err := v.isPackageConfigValid(p, activeBundle)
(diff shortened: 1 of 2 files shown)


pkg/webhook/package_webhook.go
```diff
@@ -93,6 +93,10 @@ func (v *packageValidator) isPackageConfigValid(p *v1alpha1.Package, activeBundl
 		return false, fmt.Errorf("package %s does not contain any versions", p.Name)
 	}
 
+	if packageInBundle.WorkloadOnly && !p.IsInstalledOnWorkload() {
```
ivyostosh, line 96: This won't be a blocker for this PR, but I just wonder - where do you plan to define if a package is WorkloadOnly or not and have that be in place yet?

ivyostosh (commented):
/lgtm

pkg/webhook/package_webhook.go
a-cool-train: Another PR is coming through for that.

PR #538, opened 2022-10-12
Remove personal ecr repo from tests

TerryHowe:
It really doesn't matter, but remove my personal ecr epo from repo


TerryHowe:
finally working

PR #545, opened 2022-10-13
Changing job name from test-controller to eksa-auth-refresher

junshun:
Changing name for the job that is triggered by the controller.


TerryHowe (approved):
/lgtm

PR #547, opened 2022-10-13
Remove webhook suite test

TerryHowe:
There are regular unit tests that cover the details of signature checks


a-cool-train:
> There are regular unit tests that cover the details of signature checks

But there are also package webhook tests as well in this file.

TerryHowe:
> > There are regular unit tests that cover the details of signature checks
> 
> But there are also package webhook tests as well in this file.


I'm not sure what this comment means, there are package  here tests pkg/webhook/package_webhook_test.go and bundle tests here pkg/webhook/packagebundle_test.go in this PR. There were one or two other tests in the suite, but they were duplicates of other tests.

PR #555, opened 2022-10-18
Remove manager context source from install

TerryHowe:
Remove the use of manager.source.registry from getRegistry and rename it getImageRegistry. This was created before gating existed and doesn't support that case I guess.


vincentni (approved):
/lgtm
/approve

PR #558, opened 2022-10-18
Add default registries to helm chart

TerryHowe:
Add ability to set default registries to helm chart


vincentni (commented):
I think we have one more `privateRegistry`.

TerryHowe:
> I think we have one more `privateRegistry`.

Done

vincentni (approved):
/lgtm
/approve

PR #576, opened 2022-10-20
Updating hello images remove -latest

jonahjon:

*Issue #, if available:*

*Description of changes:*

We were relying on the `-latest` functionality, but we should to use explicit tags since moving it to look at suffix.


jonathanmeier5 (approved):
/lgtm
/approve

PR #592, opened 2022-10-26
Update to oras 0.15

TerryHowe:


TerryHowe:
@jonahjon

jonahjon:
Have we tested this?

If not let me try and test all the CLI commands to make sure it doesn't break out CI

TerryHowe:
> Have we tested this?
> 
> If not let me try and test all the CLI commands to make sure it doesn't break out CI

I don't have the capability to test it completely, so basically unit tested, but not functionally

Issue #512, opened 2022-10-05
targetNamespace should be immutable in the webhook


TerryHowe:
It is my feeling that we shouldn't allow the targetNamespace to be changed for a package. The customer should uninstall and create a new package custom resource

- [x] Change CRD to add status to PBC and Package
- [x] Do deep copy for PBC and Package on reconciliation so we record previous state
- [x] Add webhook check to validate targetNamespace value and reject modifications

TerryHowe:
This may only be broken if the install succeeds. I installed two hello eks apps in default and the second failed. I edited the package config to install the failed app in another namespace and the installation succeeded

TerryHowe:
Done

Issue #202, opened 2022-05-12
Periodically reconcile the state of a package regardless of changes


TerryHowe:
We used to do this, but I think it was removed. We should periodically reconcile the state of packages regardless of their state. If someone deleted a deployment for example, we should recreate it. I'm not sure what will work and what will not work right now.

Issue #15, opened 2022-03-01
Packages deleted while the controller was down are not reconciled


lewisdiamond:
1. Install a package
2. Shut down the controller
3. Delete the package
4. Start the controller

Expected: The controller removes the installation part of the deleted package
Actual: The installation remains

TerryHowe:
Good call here. Labeling installed resources would probably help here.

TerryHowe:
This probably needs to be done first https://github.com/aws/eks-anywhere-packages/issues/107

PR #599, opened 2022-10-28
Add spec copy to package and pbc status

TerryHowe:
In order for us to tell what has changed during reconciliation, we need to keep a copy of the previous state. There are a few things for example that should be immutable like `targetNamespace` in package and we can't detect a change there unless we know the previous value.


TerryHowe:
/hold

Not for a bug release, should be December release

lewisdiamond:
We probably should keep that state elsewhere. Using the status to keep the previous spec works only for changes, not for deletes. Related to https://github.com/aws/eks-anywhere-packages/issues/15 and https://github.com/aws/eks-anywhere-packages/issues/202

TerryHowe:
> We probably should keep that state elsewhere. Using the status to keep the previous spec works only for changes, not for deletes. Related to #15 and #202


Deletes are a different problem. I assumed we were going to user a label or annotation for CR deletes while the controller is down. https://github.com/aws/eks-anywhere-packages/issues/202 is another problem not related to deletes or updates. It is about someone changing the installed package instead of the package CR

TerryHowe:
/hold

wait for this next release at this point

TerryHowe:
https://github.com/aws/eks-anywhere-packages/issues/512 depends on this for example

PR #608, opened 2022-11-04
doc(e2e): first draft for test dimensions

ewollesen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/testing/e2e-framework-design-dimensions.md
+++ b/docs/testing/e2e-framework-design-dimensions.md
@@ -1,0 +1,396 @@
+# EKS-A Curated Packages End-to-End Test Framework Design
+
+*The following is an excerpt from a larger design document that covers all aspects of curated packages end-to-end testing. These sections are provided for early comment and revision. — ewwolles@amazon.com*
+
+- [Test Variables](#org8dd3115)
+  - [Operations](#org802cd07)
+  - [Value Types](#org58512af)
+  - [Value Domains](#orgecd6ea6)
+  - [Recognized Test Dimensions](#org666546a)
+  - [Worst-case Cardinality](#orge08c449)
+  - [Implementation Design Ideas](#orgc40bdd7)
+
+<a id="#org8dd3115"></a>
+
+# Test Variables
+Many variables will define an end-to-end test run. These variables are called dimensions, and the goal of a testing system is to run tests that cover as much of the range of each dimension as possible. For some types of values, every value can be covered by tests, but for other types of values, one has to be content with some subset of values from an infinite range. The recognized dimensions for curated packages end-to-end tests are categorized and discussed below.
+
+
+<a id="org802cd07"></a>
+
+## Operations
+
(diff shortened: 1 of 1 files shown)


docs/testing/e2e-framework-design-dimensions.md
```diff
@@ -0,0 +1,396 @@
...
+| 3     | Node OSes                    |
+| 2     | Node OS versions             |
+| 2     | Packages controller versions |
+| 2     | Package versions             |
+| 1     | Architectures                |
+| 1     | Self vs Remotely managed     |
+| 1     | Bundle version               |
+| 1,536 | Total permutations           |
+
+With this far more tractable scenario, a complete run would take forty-eight minutes. Of course this still assumes 32 CPUs available and able to run any test (none of which take more than a minute from beginning to end).
```
junshun, line 320: This 48 min value is per package right?

docs/testing/e2e-framework-design-dimensions.md
```diff
@@ -0,0 +1,396 @@
...
+| Bare Metal | N/A                        |
+| CloudStack | 4.14+                      |
+| Docker     | ? (TBD)                    |
+| Snowball   | all (which is 1 right now) |
+| vSphere    | v7+ (major releases only?) |
+
+
+<a id="org1f576cb"></a>
+
+### CLI Version
```
junshun, line 142: Should the CLI even be part of this framework? It would make sense for CLI test to remain as is in the eks-anywhere and those should just handle the package installation

docs/testing/e2e-framework-design-dimensions.md
junshun: It would also be per test flow correct? For example if we there was a e2e test scenario for harbor to make sure images are imported correctly, would that need to be tested through all the permutations?

docs/testing/e2e-framework-design-dimensions.md
```diff
@@ -0,0 +1,396 @@
...
+
+This approach makes tests easy to identify, and is simple for test writers to implement. However there's added complexity in writing a tool to read, parse, and react to the go tags. The tags themselves are a bit magical, and will depend on documentation for discovery.
+
+
+<a id="orgdbecf99"></a>
+
+### Test functions to build state
+
+```go 
+func TestSomePackage(t \*testing.T) {
```
junshun, line 378: This could be a bit confusing if there are test scenarios that change between providers. Example Harbor would test differently in a dynamic provisioning environment vs not.

docs/testing/e2e-framework-design-dimensions.md
ewollesen: Yes, per test.

docs/testing/e2e-framework-design-dimensions.md
ewollesen: Presumably, the test writer would have access to some structure in which they could look up the current test dimensions under which the test was executing, and take the appropriate actions based on that.

TerryHowe:
This approach isn't flying, reopen if I'm wrong.

PR #609, opened 2022-11-08
Add ability to disable cron job

TerryHowe:


charts/eks-anywhere-packages/values.yaml
```diff
@@ -105,6 +105,7 @@ cronjob:
   repository: "ecr-token-refresher"
   # -- ECR refresher tag, or SHA256 checksum.
   tag: "{{ecr-token-refresher}}"
+  enable: true
```
junshun, line 108: Do we want the cronjob off by default?

TerryHowe: changed this

PR #610, opened 2022-11-08
Add digest to chart

TerryHowe:


ivyostosh:
/lgtm

I assume you will update build-tooling next to handle the tag replacement?

TerryHowe:
> I assume you will update build-tooling next to handle the tag replacement?

Yeh, I've already tested builds and they are working fine, but I need to merge this and tag it before I can PR that

PR #615, opened 2022-11-14
Chart.yaml version for image tag

junshun:
Changing template for controller/cronjob to use chart.yaml version value to create image url

Addresses issue where image tag doesn't exist. Currently release process updates the version in chart.yaml. This PR changes the chart to use that version to build the image url.

This should be a short-term fix until a different solution is developed to handle pulling by tag and promotion.


junshun:
/hold
Up for discussion if we want to go this path

TerryHowe (approved):
/lgtm

PR #624, opened 2022-11-17
Add createNamespace PBC setting

### LINKED ISSUE #601: Allow the package controller to create destination namesapces for Curated Package deployments
**What would you like to be added**:
Curated Packages should be allowed to create their targetNamespace perhaps with an addition to the `packages.packages.eks.amazonaws.com` CRD for `spec.createNamespace` which will default to `false` and provide the exact current behavior, but could optionally be set to `true` to tell the package controller to create the `targetNamespace` before deploying the package.

If you wanted to be extra safe, the package controller could error out if `spec.createNamespace` was set to `true` and the namespace already exists, unless the namespace was created by the package controller, which could be tracked via an annotation on the namespace.

**Why is this needed**:
Currently to deploy a Curated Package, you must first create the namespace that the package will deploy into, then you must create the Package resource in the correct eksa-packages- namespace.  This becomes even more tedious when you are deploying a Curated Package onto a workload cluster which is managed by a management cluster as you must then perform the following dance:

- change kubectl context to the target workload cluster
- create the required kubernetes namespace
- change kubectl context to the management cluster that the workload cluster belongs to
- create the requred Package resource

When deploying Curated Packages via a management tool (such as terraform) this becomes even more annoying as you must pass multiple kubernetes providers around to create namespaces on one cluster while defining packages on another cluster.

lewisdiamond:
This PBC setting will tell the controller to create the namespace on the target cluster if it doesn't exist. This is disabled by default because the controller does not impersonate the user applying the package and will always have namespace creation permissions.

Fixes #601

### DIFF
--- a/api/v1alpha1/packagebundlecontroller_types.go
+++ b/api/v1alpha1/packagebundlecontroller_types.go
@@ -85,6 +85,11 @@ type PackageBundleControllerSpec struct {
 	// Repository portion of an OCI address to the bundle
 	// +optional
 	BundleRepository string `json:"bundleRepository"`
+
+	// +kubebuilder:default:=false
+	// Allow target namespace creation by the controller
+	// +optional
+	CreateNamespace bool `json:"createNamespace"`
 }
 
 // +kubebuilder:validation:Enum=ignored;active;disconnected;upgrade available
--- a/pkg/driver/helmdriver.go
+++ b/pkg/driver/helmdriver.go
@@ -70,12 +70,12 @@ func (d *helmDriver) Initialize(ctx context.Context, clusterName string) (err er
 }
 
 func (d *helmDriver) Install(ctx context.Context,
-	name string, namespace string, source api.PackageOCISource, values map[string]interface{}) error {
+	name string, namespace string, createNamespace bool, source api.PackageOCISource, values map[string]interface{}) error {
 	var err error
 	install := action.NewInstall(d.cfg)
 	install.Version = source.Version
 	install.ReleaseName = name
-	install.Namespace = namespace
+	install.CreateNamespace = createNamespace
 
 	helmChart, err := d.getChart(install, source)
 	if err != nil {
@@ -90,6 +90,7 @@ func (d *helmDriver) Install(ctx context.Context,
 			namespace = "default"
 		}
 	}
+	install.Namespace = namespace
 
 	// Update values with imagePullSecrets
 	// If no secret values we should still continue as it could be case of public registry or local registry
--- a/pkg/packages/manager.go
+++ b/pkg/packages/manager.go
@@ -150,7 +150,9 @@ func processInstalling(mc *ManagerContext) bool {
 		mc.Log.Error(err, "Initialization failed")
 		return true
 	}
-	if err := mc.PackageDriver.Install(mc.Ctx, mc.Package.Name, mc.Package.Spec.TargetNamespace, mc.Source, values); err != nil {
+
+	createNamespace := mc.PBC.Spec.CreateNamespace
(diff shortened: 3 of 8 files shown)


pkg/driver/helmdriver.go
```diff
@@ -70,12 +70,13 @@ func (d *helmDriver) Initialize(ctx context.Context, clusterName string) (err er
...
 
 func (d *helmDriver) Install(ctx context.Context,
-	name string, namespace string, source api.PackageOCISource, values map[string]interface{}) error {
+	name string, namespace string, createNamespace bool, source api.PackageOCISource, values map[string]interface{}) error {
 	var err error
 	install := action.NewInstall(d.cfg)
 	install.Version = source.Version
 	install.ReleaseName = name
 	install.Namespace = namespace
+	install.CreateNamespace = createNamespace
```
junshun, line 79: Does helm fail installs if the namespace already exists?

api/v1alpha1/packagebundlecontroller_types.go
```diff
@@ -85,6 +85,11 @@ type PackageBundleControllerSpec struct {
 	// Repository portion of an OCI address to the bundle
 	// +optional
 	BundleRepository string `json:"bundleRepository"`
+
```
junshun, line 88: From a design perspective will we also have the responsibility to remove the namespace if we create it?

pkg/driver/helmdriver.go
```diff
@@ -70,12 +70,13 @@ func (d *helmDriver) Initialize(ctx context.Context, clusterName string) (err er
 }
 
 func (d *helmDriver) Install(ctx context.Context,
-	name string, namespace string, source api.PackageOCISource, values map[string]interface{}) error {
+	name string, namespace string, createNamespace bool, source api.PackageOCISource, values map[string]interface{}) error {
```
junshun, line 73: From a design perspective should we only allow this namespace to be created if targetNamespace is set in the Package Spec? Otherwise we would be installing in what we set as defaultNamespace in the package charts. We don't really have a way to inform the user what that defaultNamespace is now, which could lead them to be confused as to where the package actually installed

junshun: Maybe a solution here could be to generate a targetNamespace value when the Package spec is generated via the CLI

pkg/driver/helmdriver.go
lewisdiamond: No, I tested that both on a workload and management target cluster.

api/v1alpha1/packagebundlecontroller_types.go
lewisdiamond: No we should not delete the namespace that was created. It will be left there as it's currently implemented.

jonahjon:
Will this require an upgrade to the Controller, or since it's PBC it should be self update when the chart does?

pkg/packages/manager.go
```diff
@@ -150,7 +150,9 @@ func processInstalling(mc *ManagerContext) bool {
 		mc.Log.Error(err, "Initialization failed")
 		return true
 	}
-	if err := mc.PackageDriver.Install(mc.Ctx, mc.Package.Name, mc.Package.Spec.TargetNamespace, mc.Source, values); err != nil {
+
+	createNamespace := mc.PBC.Spec.CreateNamespace
```
a-cool-train, line 154: variable not necessary here?

lewisdiamond: No it's not necessary per-se but I prefer that over a function call with a bunch of deep property access.

lewisdiamond:
> Will this require an upgrade to the Controller, or since it's PBC it should be self update when the chart does?

Not sure what you mean by this. It will be a feature only available if you upgrade the controller and the CRDs. New controllers will work with existing PBCs.

TerryHowe (approved):
/lgtm

PR #657, opened 2022-11-29
Fixing the bundle diff check removing unique metadata fields

jonahjon:

*Issue #, if available:*

*Description of changes:*

Remove the metadata.Annotations and metadata.Name field before doing the diff check on the bundle files.

Tested locally using a builder-base image, and works. Applying change to Dev and Staging first to verify.


TerryHowe (approved):
/lgtm
/approve

Issue #598, opened 2022-10-27
bundle signature should exclude name of bundle


TerryHowe:
Maybe there is a reason for this, but it seems like you should be able to duplicate a bundle with a new name.

lewisdiamond:
I'd prefer keeping it part of the signature otherwise it's possible for an older bundle with packages that have known vulnerabilities to be presented as an upgrade by bumping the package name.

PR #669, opened 2022-12-06
Add bundle minimum controller requirement

lewisdiamond:
Add config package for storing configuration state (including build version)
Add validation of minimum controller version required for the bundle


api/v1alpha1/packagebundle_types.go
```diff
@@ -109,14 +110,15 @@ type PackageBundleStatus struct {
...
 type PackageBundleStateEnum string
 
 const (
-	PackageBundleStateAvailable PackageBundleStateEnum = "available"
-	PackageBundleStateIgnored   PackageBundleStateEnum = "ignored"
-	PackageBundleStateInvalid   PackageBundleStateEnum = "invalid"
+	PackageBundleStateAvailable       PackageBundleStateEnum = "available"
+	PackageBundleStateIgnored         PackageBundleStateEnum = "ignored"
+	PackageBundleStateInvalid         PackageBundleStateEnum = "invalid"
+	PackageBundleStateUpgradeRequired PackageBundleStateEnum = "upgrade required"
```
TerryHowe, line 121: This might be confusing because upgrades are tracked on the PBC.

pkg/bundle/manager.go
```diff
@@ -23,14 +25,16 @@ type bundleManager struct {
...
 }
 
-func NewBundleManager(log logr.Logger, registryClient RegistryClient, bundleClient Client, targetClient authenticator.TargetClusterClient) *bundleManager {
+func NewBundleManager(log logr.Logger, registryClient RegistryClient, bundleClient Client, targetClient authenticator.TargetClusterClient, config config.Config) *bundleManager {
 	return &bundleManager{
 		log:            log,
 		bundleClient:   bundleClient,
 		registryClient: registryClient,
 		targetClient:   targetClient,
+		config:         config,
```
TerryHowe, line 37: This config is such a generic name, I had no idea what it was until I got further down.

pkg/bundle/manager.go
```diff
@@ -66,6 +80,11 @@ func (m bundleManager) ProcessBundle(_ context.Context, newBundle *api.PackageBu
 	return false, nil
 }
 
+func (m *bundleManager) isCompatibleWith(bundle *api.PackageBundle) bool {
+	currentVersion := m.config.BuildInfo.Version
+	return currentVersion == config.DEVELOPMENT || semver.Compare(currentVersion, bundle.Spec.MinVersion) >= 0
```
TerryHowe, line 85: seems like the semver comparison belongs in the config object not to mention the DEVELOPMENT thing. The bundle manager should be blissfully ignorant.

pkg/signature/manifest.go
```diff
@@ -32,8 +32,16 @@ type GojqParams struct {
...
 var (
-	AlwaysExcluded = []string{".status", ".metadata.creationTimestamp", ".metadata.generation", ".metadata.managedFields", ".metadata.uid", ".metadata.resourceVersion"}
-	GojqTemplate   = template.Must(template.New("gojq_query").Funcs(
+	AlwaysExcluded = []string{
+		".status",
+		".metadata.creationTimestamp",
+		".metadata.generation",
+		".metadata.managedFields",
+		".metadata.uid",
+		".metadata.resourceVersion",
```
TerryHowe, line 41: what are your thoughts on adding `.metadata.name`? It would be useful to be able to duplicate a bundle.

TerryHowe (changes requested):

api/v1alpha1/packagebundle_types.go
lewisdiamond: "controller upgrade required" maybe?

pkg/bundle/manager.go
lewisdiamond: Determining if the bundle is compatible by comparing the minimum required version is part of the business logic of the bundle controller. The controller decides if it admits the bundle or not based on requirements that should not be part of a generic config package that holds application configuration.

pkg/signature/manifest.go
lewisdiamond: I actually commented a couple days back on the issue you created for this: https://github.com/aws/eks-anywhere-packages/issues/598

I think it would allow to fake upgrades which are effectively downgrades that may have CVEs. We don't really support downgrading either since we don't have downwards migrations so this could potentially make it easy for attackers to break things.

api/v1alpha1/packagebundle_types.go
TerryHowe, line 121:
```suggestion
	PackageBundleStateUpgradeRequired PackageBundleStateEnum = "controller upgrade required"
```

api/v1alpha1/packagebundle_types.go
```diff
@@ -109,14 +110,15 @@ type PackageBundleStatus struct {
 	State PackageBundleStateEnum `json:"state"`
 }
 
-// +kubebuilder:validation:Enum={"available","ignored","invalid"}
+// +kubebuilder:validation:Enum={"available","ignored","invalid","upgrade required"}
```
TerryHowe, line 113:
```suggestion
// +kubebuilder:validation:Enum={"available","ignored","invalid","controller upgrade required"}
```

TerryHowe (approved):
/lgtm
/approve

TerryHowe (approved):
/lgtm
/approve

PR #671, opened 2022-12-06
Add insecure config to bypass unknown cert issue

vincentni:


Dockerfile
```diff
@@ -1,16 +1,17 @@
 # Build the manager binary
```
TerryHowe, line 1: This docker file is not used.

pkg/artifacts/registry.go
```diff
@@ -23,7 +23,7 @@ func NewRegistryPuller() *RegistryPuller {
 }
 
 func (p *RegistryPuller) Pull(ctx context.Context, ref string) ([]byte, error) {
-	registry, err := content.NewRegistry(content.RegistryOptions{})
+	registry, err := content.NewRegistry(content.RegistryOptions{Insecure: true})
```
TerryHowe, line 26: Alarm bells are going off here. Can we pass in the cert instead?

TerryHowe (changes requested):

vincentni:
Close this PR as we are not rushing for Thursday's code freeze.

Issue #658, opened 2022-11-29
Package Support in all Regions enable/testing


jonahjon:
- [x] Create a test during the deployment which reads the versions tags/shas from every image/helm chart in the bundle and checks for their presence in ECR in the supported regions. 

- [x] Enable region parameter in token refresher.

- [ ] Create e2e test installing packages from another region.

jonahjon:
We got another issue tracking the last item here. https://github.com/aws/eks-anywhere-packages/issues/773

PR #674, opened 2022-12-07
Moving region check call to correct spot

jonahjon:

*Issue #, if available:*

#658

*Description of changes:*

The last region check was calling on the staging account for helm charts, we need to move it earlier to query vs the Account which the images live in.


TerryHowe (approved):
/lgtm
/approve

PR #697, opened 2022-12-27
Adding support for additional ecr endpoints

junshun:
*Issue #658 

*Description of changes:*
Adding support to get authorization token for all regions. List of regions comes from the service endpoints listed on https://docs.aws.amazon.com/general/latest/gr/ecr.html.

### DIFF
--- a/ecrtokenrefresher/pkg/aws/aws.go
+++ b/ecrtokenrefresher/pkg/aws/aws.go
@@ -25,20 +25,61 @@ const (
 	defaultAccountID   = "783794618700"
 	devAccountID       = "857151390494"
 	envRegionName      = "AWS_REGION"
-	regionDefault      = "us-west-2"
 )
 
-func GetECRCredentials() ([]ECRAuth, error) {
-	// Default AWS Region to us-west-2 if not set by User.
-	_, ok := os.LookupEnv(envRegionName)
-	if !ok {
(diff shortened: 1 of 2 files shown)


ecrtokenrefresher/pkg/aws/aws.go
```diff
@@ -25,20 +25,61 @@ const (
...
-			return nil, err
+			failedList = append(failedList, region)
 		}
+		creds = append(creds, regionCreds...)
+	}
+	if len(creds) < 1 {
+		return nil, failedList, fmt.Errorf("failed for all regions, %v", err)
 	}
+	return creds, failedList, nil
+}
```
TerryHowe, line 46: don't they supply the region in the aws creds?

junshun: So this is for replication of the secret in all regions. Currently the ask is to support all regions. Since copying the token and changing the endpoint doesnt work we would need to request a new token for each region.

TerryHowe: Yeh, supporting all regions to me doesn't mean supporting them all at the same time to me at least. The user would specify what region they want to use.

junshun:
Closing pr as currently there is no need to be authenticated to all locations at once due the limitations of the PBC

PR #706, opened 2022-12-30
Merge package client with manager client

TerryHowe:
The old bundle client and package client both talk to the management cluster. The package client wasn't doing much.


pkg/bundle/client.go
```diff
@@ -187,6 +193,17 @@ func (bc *managerClient) CreateClusterConfigMap(ctx context.Context, clusterName
 	return nil
 }
 
+// CreatePackage Creates the given package resource
```
ivyostosh, line 196:
```suggestion
// CreatePackage creates the given package resource
```

PR #725, opened 2023-01-06
Add secrets

vincentni:
*Description of changes:*
Add two secrets to the helm chart.

### DIFF
--- a/charts/eks-anywhere-packages/templates/cronjob.yaml
+++ b/charts/eks-anywhere-packages/templates/cronjob.yaml
@@ -1,4 +1,4 @@
-#{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version }}
+{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version }}
 apiVersion: batch/v1
 {{- else }}
 apiVersion: batch/v1beta1
@@ -29,12 +29,12 @@ spec:
                   valueFrom:
                     secretKeyRef:
                       name: aws-secret
-                      key: ID
+                      key: AWS_ACCESS_KEY_ID
                 - name: AWS_SECRET_ACCESS_KEY
                   valueFrom:
                     secretKeyRef:
                       name: aws-secret
-                      key: SECRET
+                      key: AWS_SECRET_ACCESS_KEY
                 - name: AWS_REGION
                   valueFrom:
                     secretKeyRef:
--- a/charts/eks-anywhere-packages/templates/cronjob.yaml
+++ b/charts/eks-anywhere-packages/templates/cronjob.yaml
@@ -1,4 +1,4 @@
-{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version }}
+#{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version }}
 apiVersion: batch/v1
 {{- else }}
 apiVersion: batch/v1beta1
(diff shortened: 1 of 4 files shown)


charts/eks-anywhere-packages/templates/cronjob.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version }}
+#{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version }}
```
TerryHowe, line 1: what is this change?

charts/eks-anywhere-packages/templates/registrymirrorsecret.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+data:
+  {{- with .Values.registryMirrorSecret }}
+  ENDPOINT: "{{ .endpoint | b64enc }}"
+  USERNAME: "{{ .username | b64enc }}"
+  PASSWORD: "{{ .password | b64enc }}"
+  CACERTCONTENT: "{{ .cacertcontent | b64enc }}"
```
TerryHowe, line 7: These should be base64 encoded when they are put in the values.yaml or we are going to have problems.

TerryHowe: On the cert, I think it makes sense to base64 encode it for the newlines rather than do a special space replace.

junshun:
The concern with this design is how are expecting these values to be based in? I assume they will be passed in already base64 encoded. Will this be done via the CLI during helm install? If so we will need to update the logging in the CLI as I believe with helm install the values do show up in the log.

vincentni:
> The concern with this design is how are expecting these values to be based in? I assume they will be passed in already base64 encoded. Will this be done via the CLI during helm install? If so we will need to update the logging in the CLI as I believe with helm install the values do show up in the log.

Yes, CLI does the encoding and stores them into a temporary file which helm install may read. The logging in the CLI only shows that temporary file name, not those encoded values.

charts/eks-anywhere-packages/templates/cronjob.yaml
```diff
@@ -29,20 +29,37 @@ spec:
...
-                      key: SECRET
-                      optional: true
+                      key: AWS_SECRET_ACCESS_KEY
                 - name: AWS_REGION
                   valueFrom:
                     secretKeyRef:
                       name: aws-secret
                       key: REGION
-                      optional: true
+                - name: REGISTRY_MIRROR_ENDPOINT
```
jonahjon, line 43: Should these be optional values? Or are we ok with empty strings when nothing is passed.

charts/eks-anywhere-packages/templates/registrymirrorsecret.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+data:
```
jonahjon, line 2: nit, but the ordering of this is wierd.

Usually Data is after kind, and metadata fields when reading k8s files.

charts/eks-anywhere-packages/templates/cronjob.yaml
vincentni: The secret gets created even though its fields are all empty, which allows customers to edit those secrets when they want to use such authentication.

charts/eks-anywhere-packages/templates/registrymirrorsecret.yaml
vincentni: reordered them. thanks!

TerryHowe (approved):
/lgtm

PR #736, opened 2023-01-11
Adding registry mirror secrets to all namespaces

vincentni:


charts/eks-anywhere-packages/templates/deployment.yaml
```diff
@@ -66,6 +66,11 @@ spec:
               value: {{ .Values.proxy.NO_PROXY | quote}}
             - name: CLUSTER_NAME
               value: {{ .Values.clusterName | quote}}
```
TerryHowe, line 68: Why isn't this a file? This doesn't seem to be used.

charts/eks-anywhere-packages/templates/deployment.yaml
```diff
@@ -90,11 +95,18 @@ spec:
...
             name: cert
             readOnly: true
+          - mountPath: /tmp/registry-mirror
+            name: cert
+            readOnly: true
       volumes:
       - name: cert
         secret:
           defaultMode: 420
           secretName: webhook-server-cert
```
TerryHowe, line 105: Seems like the mode should be 0200 for these things, but not a huge deal. Also odd to me to use the /tmp directory, but that is what the webhooks are doing. It does seem like a problem that the volume mount name is the same for webhooks and the registsry.

charts/eks-anywhere-packages/templates/deployment.yaml
TerryHowe: Not sure if REGISTRY_MIRROR in general is a good name, could be a private registry even in ECR

TerryHowe (approved):
Steven should probably look at this, but I have no issue

ecrtokenrefresher/pkg/secrets/aws/aws.go
```diff
@@ -0,0 +1,185 @@
+package aws
```
junshun, line 1: Would be good to do a git mv on this file from the previous so it just shows up as changed/keeps history

ecrtokenrefresher/pkg/secrets/registrymirror/registrymirror.go
```diff
@@ -0,0 +1,88 @@
+package registrymirror
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/aws/eks-anywhere-packages/ecrtokenrefresher/pkg/constants"
```
junshun, line 7: Import statement order should be fix, suprised this wasn't caught by golang-ci

ecrtokenrefresher/pkg/secrets/registrymirror/registrymirror.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (mirror *RegistryMirrorSecret) IsActive() bool {
+	if val, _ := os.LookupEnv(usernameEnv); val == "" {
+		return false
+	}
+	if val, _ := os.LookupEnv(passwordEnv); val == "" {
+		return false
+	}
+	return true
```
junshun, line 46: Should check endpoint here as well or else it would continue even with no endpoint

charts/eks-anywhere-packages/templates/job.yaml
```diff
@@ -0,0 +1,58 @@
...
+metadata:
+  name: init-job
+  namespace: eksa-packages
+spec:
+  template:
+    spec:
+      restartPolicy: OnFailure
+      serviceAccountName: {{ include "eks-anywhere-packages.serviceAccountName" . }}
+      containers:
+      - name: init-cronjob
```
junshun, line 12: Change this name to init-job, this run doesn't have anything to do with the cronjob itself

charts/eks-anywhere-packages/templates/job.yaml
```diff
@@ -0,0 +1,58 @@
...
+        image: {{.Values.sourceRegistry}}{{ template "template.image" .Values.cronjob }}
+        imagePullPolicy: {{ .Values.imagePullPolicy }}
+        env:
+          - name: ECR_TOKEN_SECRET_NAME
+            value: ecr-token
+          - name: AWS_ACCESS_KEY_ID
+            valueFrom:
+              secretKeyRef:
+                name: aws-secret
+                key: AWS_ACCESS_KEY_ID
```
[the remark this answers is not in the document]
vincentni: They used to be not part of the helm chart, so it should be fine. Helm complaints anyway as long as the resource with the same name already exists.

ecrtokenrefresher/pkg/secrets/registrymirror/registrymirror.go
vincentni: I ran golang-ci locally but didn't get complaints about that. I also ran gofmt to format all code.

ecrtokenrefresher/pkg/secrets/aws/aws.go
vincentni: noted.

PR #740, opened 2023-01-13
Bring over the pull function

TerryHowe:
Bring over the pull function to download bundles. Need to read the manifest to get the sha of the bundle.yaml file.


pkg/artifacts/registry.go
```diff
@@ -23,21 +23,29 @@ func NewRegistryPuller() *RegistryPuller {
...
+		return nil, err
 	}
-	store := content.NewMemory()
 
-	_, err = oras.Copy(ctx, registry, ref, store, "")
+	certificates, err := registry.GetCertificates("registry.crt")
 	if err != nil {
-		return nil, fmt.Errorf("pulling artifact %q: %s", ref, err)
+		return nil, err
 	}
```
vincentni, line 35: What if no cert is required like pulling from private ecr or insecure local registry? I think we need a way to know if cert is intended to be used and errors out accordingly.

TerryHowe: This will return an empty X509 struct if the file is empty which will be the case in the pod if the volume secret is mounted. I'm not a fan of completely ignoring the error though.

TerryHowe: okay, I passed a logger down, so it can just log the error and ignore it. I also updated the name of the file to match the helm chart

pkg/artifacts/registry.go
```diff
@@ -2,42 +2,57 @@ package artifacts
...
-		return nil, fmt.Errorf("creating registry: %w", err)
+		return nil, err
 	}
-	store := content.NewMemory()
 
-	_, err = oras.Copy(ctx, registry, ref, store, "")
+	certificates, err := registry.GetCertificates(certFile)
 	if err != nil {
-		return nil, fmt.Errorf("pulling artifact %q: %s", ref, err)
+		p.log.Error(err, "problem getting certificate file", "filename", certFile)
```
vincentni, line 41: shall we return from this method right after logging?

TerryHowe: No, best effort if there is no file. We don't need certs for ECR for example, so that will still work.

pkg/artifacts/registry.go
```diff
@@ -2,42 +2,57 @@ package artifacts
...
-	"fmt"
 
-	"oras.land/oras-go/pkg/content"
-	"oras.land/oras-go/pkg/oras"
+	"github.com/go-logr/logr"
+
+	"github.com/aws/eks-anywhere-packages/pkg/registry"
 )
 
+const certFile = "/tmp/registry-mirror/CACERTCONTENT"
```
a-cool-train, line 11: We need to read this from a secret config, right?

TerryHowe: The secret is mounted on /tmp/registry-mirror as a file

vincentni: It has been updated to /tmp/config/registry/ca.crt

TerryHowe: I don't think it is a good idea to use a copy when we can use the original. This version is a lot more likely to be correct.

vincentni (approved):
/lgtm
/approve

PR #746, opened 2023-01-16
Reduce path for controller-gen

TerryHowe:
The generatebundle command causes controller-gen to generate the file `config/crd/bases/_.yaml` since the generate bundle uses  kubebuilder style structs. This stops the generation of that useless file.


ewollesen (commented):
/lgtm

PR #747, opened 2023-01-17
Feature helm ca trust

a-cool-train:
*Description of changes:*
- Currently, helm doesn't work on self signed certs. Adding this as a fix to resolve this issue.
- Previously, curated packages didn't support pulling from a registry with custom CA, these PR fixes this issue for helm.


pkg/driver/helmdriver.go
```diff
@@ -243,3 +246,43 @@ func (d *helmDriver) IsConfigChanged(_ context.Context, name string, values map[
 
 	return !reflect.DeepEqual(values, rel.Config), nil
 }
+
```
TerryHowe, line 249: Do you need a note that these are just copy/pasta and can be removed when helm releases this stuff?

pkg/driver/helmdriver.go
```diff
@@ -35,6 +35,10 @@ type helmDriver struct {
 
 var _ PackageDriver = (*helmDriver)(nil)
 
+// TODO: Temporarily hard coding but needs to be read from secrets
+var caFile = "/tmp/registry-mirror/CACERTCONTENT"
```
vincentni, line 39: It is now at /tmp/config/registry/ca.crt

ivyostosh:
Presubmit failed

pkg/driver/helmdriver.go
TerryHowe: If there is any change to CACERTCONTENT, cat.crt won't be updated for something like 12 hours. It is not a good idea to copy that secret.

TerryHowe:
That error on presubmit is because there is a mismatch between go.mod top level and generate bundle. I think just need to run `make build` down there or `go mod tidy`

pkg/driver/helmdriver.go
a-cool-train: With the current PR, this change is needed regardless of whether helm releases the fix because these changes are done on the CLI and not the main path so mimicking this in our repo is required.

TerryHowe:
/hold

Need to investigate if we can do without this change

pkg/driver/helmdriver.go
```diff
@@ -35,6 +35,10 @@ type helmDriver struct {
 
 var _ PackageDriver = (*helmDriver)(nil)
 
+// TODO: Temporarily hard coding but needs to be read from secrets
+var caFile = "/tmp/config/registry/CACERTCONTENT"
+var insecure = false
```
vincentni, line 40: insecure flag is in the registrymirrorsecret.yaml, you can add that as an env variable to deployment.yaml and then read from there

pkg/driver/helmdriver.go
```diff
@@ -35,6 +35,10 @@ type helmDriver struct {
 
 var _ PackageDriver = (*helmDriver)(nil)
 
+// TODO: Temporarily hard coding but needs to be read from secrets
+var caFile = "/tmp/config/registry/CACERTCONTENT"
```
vincentni, line 39: update that to `/tmp/config/registry/ca.crt`

a-cool-train: Done

pkg/driver/helmdriver.go
```diff
@@ -35,6 +37,11 @@ type helmDriver struct {
 
 var _ PackageDriver = (*helmDriver)(nil)
 
+var (
+	caFile         = "/tmp/config/registry/ca.crt"
+	insecureEnvVar = "REGISTRY_INSECURE"
```
vincentni, line 42: that should be a constant

vincentni (approved):
/lgtm
/approve
/unhold

PR #756, opened 2023-01-17
Rework SetURI into ParseArtifactFromURI

ewollesen:
By moving SetURI to ParseArtifactFromURI we re-use commonly used terminology and semantics around the term "Parse", which implies that the process can fail (returning nil and an error), as opposed to the SetURI, which left it up to the caller to initialize the Artifact, and abort if the SetURI call failed.

Also adds some tests for the parsing algorithm, and changes methods to remove the pointer receiver which better communicates that the methods are read-only.


TerryHowe (approved):
/lgtm
/approve

Thanks, I had a note to add tests for that.

PR #757, opened 2023-01-18
Lift configfile

ewollesen:
Two smallish changes:

    Lift configfile.ConfigFile out of CredentialsStore
    
    Rather than assuming that users of CredentialStore will know to call Init
    after instantiation, the CredentialStore's dependency on a Docker *ConfigFile
    is made explicit by requiring it to be passed to the constructor. Now we need
    never worry that a user will forget to call Init, and it removes the awkward
    pattern of New-then-Init.
 

and:

    Don't store the RegistryPuller's StorageClient
    
    There's no point in storing it, as it's never reused.

### DIFF
--- a/pkg/registry/credentials.go
+++ b/pkg/registry/credentials.go
@@ -1,44 +1,28 @@
 package registry
 
 import (
-	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/config/credentials"
 	"oras.land/oras-go/v2/registry/remote/auth"
 )
 
-// CredentialStore for registry credentials such as ~/.docker/config.json.
-type CredentialStore struct {
-	directory  string
+// DockerCredentialStore for Docker registry credentials, like ~/.docker/config.json.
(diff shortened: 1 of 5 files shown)


jonahjon:
/lgtm

Looks like a good refactor

pkg/registry/credentials.go
```diff
@@ -1,44 +1,28 @@
...
-	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/config/credentials"
 	"oras.land/oras-go/v2/registry/remote/auth"
 )
 
-// CredentialStore for registry credentials such as ~/.docker/config.json.
-type CredentialStore struct {
-	directory  string
+// DockerCredentialStore for Docker registry credentials, like ~/.docker/config.json.
```
TerryHowe, line 9: This is all fine, but I don't like docker in the name. I was trying to hide that. We don't have to be stuck with docker forever.

vincentni:
We may want to revisit this PR as bundle pulling failed due to 401 unauthorized access.

PR #766, opened 2023-01-19
Add tag option for package controller dev promote pipeline

jonahjon:

*Issue #, if available:*

*Description of changes:*

Why this needs to exist is because there is two sets of dev build being produced, one by the EKS-A pipeline using Git Tag, and another by the new dev pipeline. The current logic just defaults to latest, so adding a tag option allows us to always promote our dev builds for e2e purposes to always use dev chart.

### DIFF
--- a/generatebundlefile/ecr_helper.go
+++ b/generatebundlefile/ecr_helper.go
@@ -106,7 +106,7 @@ func ImageTagFilter(details []ImageDetailsBothECR, version string) []ImageDetail
 	var filteredDetails []ImageDetailsBothECR
 	for _, detail := range details {
 		for _, tag := range detail.ImageTags {
-			if strings.Contains(tag, version) {
+			if strings.HasPrefix(tag, version) {
 				filteredDetails = append(filteredDetails, detail)
 			}
 		}
--- a/generatebundlefile/main.go
+++ b/generatebundlefile/main.go
@@ -91,9 +91,13 @@ func cmdPromote(opts *Options) error {
 
 	promoteCharts := make(map[string][]string)
 
+	if opts.tag != "" {
+		opts.tag = "latest"
+	}
+
 	// If we are promoting an individual chart with the --promote flag like we do for most charts.
 	if opts.promote != "" {
-		promoteCharts[opts.promote] = append(promoteCharts[opts.promote], "latest")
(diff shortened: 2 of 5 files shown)


generatebundlefile/main.go
```diff
@@ -91,9 +91,13 @@ func cmdPromote(opts *Options) error {
 
 	promoteCharts := make(map[string][]string)
```
TerryHowe, line 93: this doesn't make much sense

jonahjon: What doesn't make sense the `promoteCharts`?

jonahjon: It was added for scenario's where we are promoting multiple charts for a single pipeline like cluster-autoscaler, metrics server, emissary, metallb and we are using an input file to specify which ones to promote.

TerryHowe: well, if you pass in a tag, it gets set to latest. This is either wrong or the tag should be a boolean type

jonahjon: OOoooooooo, yea I did the logic backwards let me fix that.

jonahjon: Gonna open a new PR with the fix, and the script changes actually.

generatebundlefile/ecr_helper.go
```diff
@@ -106,7 +106,7 @@ func ImageTagFilter(details []ImageDetailsBothECR, version string) []ImageDetail
 	var filteredDetails []ImageDetailsBothECR
 	for _, detail := range details {
 		for _, tag := range detail.ImageTags {
```
TerryHowe, line 108: Our bundles are like `v1-24-latest` is that going to be a problem with this?

jonahjon: No it wouldn't be a problem with that, this is only for promoting helm chart, so it's not using the k8s version in this logic at all.

I have it working locally, just working on adding some tests right now.

jonahjon:
/hold
working on tests

generatebundlefile/main.go
jonahjon: Fixed

TerryHowe (approved):
/lgtm

PR #771, opened 2023-01-22
Add tag to bundle images

TerryHowe:
The VersionImages object does not have a version. When we copy the bundle to a private registry, it would be convenient to have one to make tagging the images easy.

### DIFF
--- a/api/testdata/bundle_one.yaml
+++ b/api/testdata/bundle_one.yaml
@@ -8,12 +8,15 @@ spec:
   packages:
     - name: hello-eks-anywhere
       source:
-        registry: public.ecr.aws/l0g8r8j6
         repository: hello-eks-anywhere
         versions:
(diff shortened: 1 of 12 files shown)


jonahjon:
/lgtm

I'll want to verify this is still working just fine with bundle creation, as the tests for it are not triggered here although it relies on those structs.

lewisdiamond:
Can you add a test case (could simply modify one of the test bundle) that adds an entry in `VersionImages` but has no `Version` defined?

TerryHowe:
It appears the signing script does not generate the correct signature when the version is not specified.

jonahjon:
although you might have to manually merge this due to the test failures.

TerryHowe:
Closes: https://github.com/aws/eks-anywhere-packages/issues/799

Issue #720, opened 2023-01-04
update package controller helm chart to create secrets


vincentni:

TerryHowe:
Done

PR #778, opened 2023-01-23
Add workloadOnly field to helm chart

vincentni:
The new field governs what resource is created for management clusters and workload clusters

### DIFF
--- a/charts/eks-anywhere-packages/templates/packagebundlecontroller.yaml
+++ b/charts/eks-anywhere-packages/templates/packagebundlecontroller.yaml
@@ -4,7 +4,11 @@ apiVersion: packages.eks.amazonaws.com/v1alpha1
 kind: PackageBundleController
 metadata:
   name: {{.Values.clusterName}}
+  {{- if not .Values.workloadOnly }}
   namespace: eksa-packages
+  {{- else }}
+  namespace: eksa-packages-{{ .Values.clusterName }}
+  {{- end }}
 spec:
   upgradeCheckInterval: 24h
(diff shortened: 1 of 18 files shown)


TerryHowe:
when we install this, do we use a different name? Like `eks-anywhere-packages-clusterName`?

TerryHowe:
/hold

Just so people can see this before it merges

jonahjon:
So would this just install a helm chart with just the CRDs?

TerryHowe:
> So would this just install a helm chart with just the CRDs?

The CRDs will already be there.

TerryHowe:
Closes: https://github.com/aws/eks-anywhere-packages/issues/720

TerryHowe:
Is there going to be a follow-up with the secret?

TerryHowe (approved):
/lgtm

charts/eks-anywhere-packages/templates/packagebundlecontroller.yaml
```diff
@@ -4,7 +4,11 @@ apiVersion: packages.eks.amazonaws.com/v1alpha1
 kind: PackageBundleController
 metadata:
   name: {{.Values.clusterName}}
+  {{- if not .Values.workloadOnly }}
   namespace: eksa-packages
```
a-cool-train, line 8: Namespace for management cluster also has the clustername on it.

lewisdiamond:
I think that's fine for now but it highlights the need for a leaner process to push out charts. This should be its own chart given how it excludes almost everything. Given the time constraints we'll go ahead as-is.

PR #781, opened 2023-01-25
Fix controller service targetPort

ivyostosh:
*Issue #, if available:* https://github.com/aws/eks-anywhere-packages/issues/780


TerryHowe (approved):
/lgtm

I almost feel like it is supposed to use something like /etc/services to find it and we just have a mismaatch

PR #792, opened 2023-01-31
Allow each workload cluster to have its own registry mirror secret

vincentni:


charts/eks-anywhere-packages/templates/deployment.yaml
```diff
@@ -126,26 +126,6 @@ spec:
...
-          - name: REGISTRY_MIRROR_PASSWORD
-            valueFrom:
-              secretKeyRef:
-                name: registry-mirror-secret
-                key: PASSWORD
-          - name: REGISTRY_MIRROR_CACERTCONTENT
-            valueFrom:
-              secretKeyRef:
-                name: registry-mirror-secret
-                key: CACERTCONTENT
```
TerryHowe, line 148: 👍

charts/eks-anywhere-packages/templates/registrymirrorsecret.yaml
```diff
@@ -13,7 +13,6 @@ data:
   USERNAME: "{{ .username }}"
   PASSWORD: "{{ .password }}"
   CACERTCONTENT: "{{ .cacertcontent }}"
-  CERTCONTENT: "{{ .certcontent }}"
```
TerryHowe, line 16: 👍

TerryHowe (commented):
/lgtm

charts/eks-anywhere-packages/templates/awssecret.yaml
```diff
@@ -1,16 +1,14 @@
+{{- if not .Values.workloadOnly }}
```
junshun, line 1: Will adding this change how helm updates work?

TerryHowe: I don't think so, I think that would depend on the annotation that Lewis was going to add. This doesn't really help or hurt that.

ecrtokenrefresher/pkg/secrets/registrymirror/registrymirror.go
```diff
@@ -15,78 +14,109 @@ import (
...
-func (mirror *RegistryMirrorSecret) GetCredentials() ([]*secrets.Credential, error) {
-	utils.InfoLogger.Println("fetching auth data from Registry Mirror... ")
-	secrets := []*secrets.Credential{
-		{
-			Registry: os.Getenv(endpointEnv),
-			Username: os.Getenv(usernameEnv),
-			Password: os.Getenv(passwordEnv),
-		},
+func (mirror *RegistryMirrorSecret) GetName() string {
+	return "registry-mirror-secret"
```
lewisdiamond, line 49: Similar to above.

charts/eks-anywhere-packages/templates/awssecret.yaml
lewisdiamond: Was this secret being used when created in namespaces other than the management one?

charts/eks-anywhere-packages/values.yaml
```diff
@@ -120,7 +120,6 @@ registryMirrorSecret:
   username: ""
   password: ""
   cacertcontent: ""
-  certcontent: ""
   insecure: "ZmFsc2UK"
```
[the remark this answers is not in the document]
TerryHowe: that is `false` base64 encoded

ecrtokenrefresher/pkg/secrets/aws/aws.go
```diff
@@ -74,7 +75,23 @@ func (aws *AwsSecret) IsActive() bool {
 	return true
 }
 
-func (aws *AwsSecret) GetCredentials() ([]*secrets.Credential, error) {
+func (aws *AwsSecret) GetName() string {
+	return "aws-secret"
```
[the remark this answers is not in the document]
TerryHowe: I'm not sure if you are proposing we change the name of the secret and I think it is a little late for that.

I agree it appears it should be a public const. The only reason to do it this way is if it was implementing an interface and this was part of the interface.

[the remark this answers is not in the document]
lewisdiamond: I mean this.
```suggestion
	return aws.secretName
```

charts/eks-anywhere-packages/values.yaml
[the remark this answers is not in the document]
lewisdiamond: We should consider using the `b64enc` helm function instead of pre-encoding in the values.

charts/eks-anywhere-packages/templates/awssecret.yaml
lewisdiamond: After talking with Vincent, the previous version was never released so there shouldn't be any problem.

ecrtokenrefresher/pkg/secrets/aws/aws.go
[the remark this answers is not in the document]
TerryHowe: There are clusters in production already with the name "aws-secret". We would have to migrate that. I see no value in that.

ecrtokenrefresher/pkg/secrets/registrymirror/registrymirror.go
lewisdiamond:
```suggestion
	return mirror.credName
```

charts/eks-anywhere-packages/values.yaml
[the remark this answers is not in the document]
vincentni: The tricky part is the CACERT which is a multi-line string, and that can cause a lot of trouble when dumping it to a temporary values.yaml and then re-read and encode it from there.

PR #793, opened 2023-01-31
Adding priority to tag substring promote cmd

jonahjon:

*Description of changes:*

When promoting images, some images are doubletagged, and this will help ensure it preserves the correct tag during the copy command that was specified by the promote command.

ecr-token-refresher has tags `0.0.0-abc` and `release-0.14`.

If we use promote with substring match for `0.0.0` we want to return that tag back to the copy command, and not the `release-0.14` one since both get returned by the sdk query.


TerryHowe (approved):
/lgtm

PR #800, opened 2023-02-01
Bumping cluster autoscaler tag

junshun:
Bumping autoscaler tags to production


junshun:
/hold
Waiting for suppression rule updates

ivyostosh:
You need to update https://github.com/aws/eks-anywhere-packages/blob/main/generatebundlefile/data/prod_artifact_move.yaml and https://github.com/aws/eks-anywhere-packages/blob/main/generatebundlefile/data/staging_artifact_move.yaml as well

junshun:
done
/unhold

PR #803, opened 2023-02-02
Remove unused configmap

vincentni:
The `ns-secret-map` configMap in `eksa-packages` namespace is no longer needed as package controller is now responsible for creating that in the target namespace in form of `eksa-packages-[clusterName]` (https://github.com/aws/eks-anywhere-packages/blob/main/pkg/bundle/client.go#L162). We are no longer supporting for package installation in `eksa-packages` namespace.


TerryHowe (approved):
/lgtm

PR #806, opened 2023-02-03
Adjustments to local build to work with dev bundles

jonahjon:

*Issue #, if available:*

*Description of changes:*

Makes the make run command with Dev Bundles correctly for easier local development.

Included a screenshot showing how with the new commands you run the local server, and immediately begin installing packages from dev bundles.

Screen Shot 2023-02-03 at 4 17 13 PM

### DIFF
--- a/config/packagebundlecontroller/pbc.yaml.tmp
+++ b/config/packagebundlecontroller/pbc.yaml.tmp
@@ -1,0 +1,12 @@
+apiVersion: packages.eks.amazonaws.com/v1alpha1
+kind: PackageBundleController
+metadata:
+  name: ${CLUSTER_NAME}
+  namespace: eksa-packages
+spec:
+  activeBundle: ${BUNDLE}
+  upgradeCheckInterval: 24h
+  upgradeCheckShortInterval: 5m
+  defaultRegistry: public.ecr.aws/l0g8r8j6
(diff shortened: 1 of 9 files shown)


config/token/cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+                  name: aws-secret
+            - name: AWS_REGION
+              valueFrom:
+                secretKeyRef:
+                  key: REGION
+                  name: aws-secret
+            - name: HTTP_PROXY
+            - name: HTTPS_PROXY
+            - name: NO_PROXY
+            image: public.ecr.aws/l0g8r8j6/ecr-token-refresher:v0.0.0-f214683ead9b2ba32459d268ce9151f991375c6f
```
junshun, line 35: Is there a way for this to run the latest image instead? Or to pull the values for this from the chart? This is useful if changes are made in ecr-token-refresher or in a case where the chart itself is updated.

config/packagebundlecontroller/pbc.yaml.tmp
```diff
@@ -0,0 +1,12 @@
+apiVersion: packages.eks.amazonaws.com/v1alpha1
```
TerryHowe, line 1: Can we use the helm chart for this and pass in values.yaml with the image tages?

TerryHowe:
Too much effort at this point

PR #809, opened 2023-02-08
Add a new chart (eks-anywhere-packages-crds)

lewisdiamond:
This chart contains the CRDs for eks-anywhere-packages. 
Will be used with self-managed package manager


TerryHowe (approved):
/lgtm

PR #817, opened 2023-02-10
Fix when new version bundle doesn't exist yet

jonahjon:

In case a new bundle exist, just create an empty file for comparison so the new bundle will always get pushed.


vincentni (approved):
/lgtm
/approve

PR #828, opened 2023-02-20
Using private ECR tag not public

jonahjon:
*Issue #, if available:*

*Description of changes:*

We need to use the corresponding Private ECR tag, not the public one for the copy helm charts.


vincentni (approved):
/lgtm
/approve

PR #831, opened 2023-02-22
Make the package controller a package

lewisdiamond:
*Description of changes:*
Helm chart changes for package controller as a package. This removes the CRDs from the main package meaning this will break installation with the current CLI. The CLI must be updated once this is available.


lewisdiamond:
/hold
I want to test keeping the CRDs in the main chart temporarily.

charts/eks-anywhere-packages/templates/package.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if not .Values.workloadOnly }}
+{{-  $namespace := printf "%s-%s" "eksa-packages" .Values.clusterName -}}
+{{- if not (lookup "packages.eks.amazonaws.com/v1alpha1" "Package" "eksa-packages-main" "eks-anywhere-packages") }}
+apiVersion: packages.eks.amazonaws.com/v1alpha1
+kind: Package
+metadata:
+  annotations:
+    "helm.sh/resource-policy": keep
+  name: eks-anywhere-packages
```
jonahjon, line 9: So moving the namespace to the CRDs would make it so upgrading itself wouldn't fail right?

PR #835, opened 2023-02-24
Create packages namespaces via helm

ewollesen:
Previously, the helm chart created a specific namespace per cluster:

Self-managed clusters: "eksa-packages"
Workload clusters:     "eksa-packages-$CLUSTER_NAME"

The PBC controller was also checking for, and creating if necessary, "eksa-packages-$CLUSTER_NAME", which would only trigger for self-managed clusters, as the helm chart was creating it for workload clusters.

With this change, both "eksa-packages" and "eksa-packages-$CLUSTER_NAME" are created via the helm chart. The code to create them from the PBC controller is removed. This should simplify the entire process.

*Issue #, if available:*

*Description of changes:*


charts/eks-anywhere-packages/templates/namespace.yaml
```diff
@@ -1,13 +1,21 @@
...
 apiVersion: v1
 kind: Namespace
 metadata:
   annotations:
     "helm.sh/resource-policy": keep
   creationTimestamp: null
-  {{- if not .Values.workloadOnly }}
   name: eksa-packages
-  {{- else }}
+spec: {}
```
lewisdiamond, line 9: I'd remove the spec and status if empty.

ewollesen: Done. I didn't add those, but sure, can't hurt to remove them. Just extra noise in the review.

TerryHowe (approved):
/lgtm

vincentni (approved):

vincentni (approved):
/lgtm
/approve

charts/eks-anywhere-packages/templates/namespace.yaml
```diff
@@ -1,10 +1,17 @@
-{{ $namespace := ternary "eksa-packages" (printf "%s-%s" "eksa-packages" .Values.clusterName) .Values.workloadOnly }}
-{{- if not (lookup "v1" "Namespace" "" $namespace ) }}
+{{- if not .Values.workloadOnly }}
```
lewisdiamond, line 1: I don't think this change is necessary considering the latest version.

ewollesen:
After some discussion with @lewisdiamond we think we've come to a good arrangement that will work now, and will ease the transition to packages as a package that is to come shortly. There were some subtle edge cases he's seen that we can avoid, at the minor cost of a little extra complexity in this PR.

I'll be pushing a new commit shortly.

PR #842, opened 2023-02-27
Add more compatibility cases for helm chart namespace

lewisdiamond:
This will fail without https://github.com/aws/eks-anywhere-packages/pull/835 as it creates the cluster namespaces in the chart.

### DIFF
--- a/charts/eks-anywhere-packages/templates/namespace.yaml
+++ b/charts/eks-anywhere-packages/templates/namespace.yaml
@@ -1,17 +1,37 @@
+# If workloadOnly is true, the eksa-packages namespace isn't used and already
+# exists.
 {{- if not .Values.workloadOnly }}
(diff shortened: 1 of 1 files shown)


charts/eks-anywhere-packages/templates/namespace.yaml
```diff
@@ -1,17 +1,37 @@
...
-  creationTimestamp: null
-  name: eksa-packages
+  name: {{ $namespace }}
 ---
-{{ end }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{ $pkgnamespace := (printf "%s-%s" "eksa-packages" .Values.clusterName) }}
```
ewollesen, line 30: I don't really like the creation of this variable for no good reason, but not a blocker.

PR #852, opened 2023-03-03
Copying images over for packages dev chart for e2e

jonahjon:
*Issue #, if available:*

*Description of changes:*

When we added a promote file for the Package controller as a package, it didn't include the `copy-images` flag which is needed for the e2e dev chart to have it's image in the public ECR. Adding it back here.


TerryHowe (approved):
/lgtm

PR #853, opened 2023-03-03
Avoid one hour wait for pbc

TerryHowe:
Several errors happened during start-up and none of them matched the "connection refused", but they did result in a one hour timer for the PBC.

### DIFF
--- a/controllers/packagebundlecontroller_controller.go
+++ b/controllers/packagebundlecontroller_controller.go
@@ -17,7 +17,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


controllers/packagebundlecontroller_controller.go
```diff
@@ -127,8 +126,7 @@ func (r *PackageBundleControllerReconciler) Reconcile(ctx context.Context, req c
 
 	err = r.bundleManager.ProcessBundleController(ctx, pbc)
 	if err != nil {
-		if strings.Contains(err.Error(), "connect: connection refused") &&
-			!r.webhookInitialized {
+		if !r.webhookInitialized {
```
lewisdiamond, line 129:
```suggestion
		if strings.Contains(err.Error(), "connect: connection refused") || !r.webhookInitialized {
```

TerryHowe: This is only looking for new bundles, so I think if we have a bundle an hour wait is fine to check for an update.

PR #854, opened 2023-03-03
Up the logging for package controller

TerryHowe:
Level 6 logging gets all the appropriate data and does not overwhelm.


ewollesen:
I don't mind this at all, but it'd be preferable for us to be able to set a logging level via the e2e test and have it passed through to all these places... I know that's a pretty involved task, because it touches lots of things, but it would make situations like this a lot easier to deal with. :)

PR #858, opened 2023-03-07
Add image pull secrets to migrations

lewisdiamond:
*Description of changes:*
Add pull secret to migrations since the sourceRegistry set by the controller points to private ECR. This might change soon by passing values down to dependencies but having it available is still a good thing.

### DIFF
--- a/charts/eks-anywhere-packages-migrations/templates/adopt-crds-awssecret.yaml
+++ b/charts/eks-anywhere-packages-migrations/templates/adopt-crds-awssecret.yaml
@@ -58,6 +58,10 @@ spec:
     spec:
       restartPolicy: Never
       serviceAccountName: {{ .Values.serviceAccount.name }}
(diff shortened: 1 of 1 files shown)


charts/eks-anywhere-packages-migrations/templates/adopt-crds-awssecret.yaml
```diff
@@ -58,6 +58,10 @@ spec:
     spec:
       restartPolicy: Never
       serviceAccountName: {{ .Values.serviceAccount.name }}
```
TerryHowe, line 60: This isn't in the values.yaml how would someone know to add this?

lewisdiamond: (note that this might not even be required if we release the oci auth package at the same time)

PR #863, opened 2023-03-08
allow eksa-controller-manager management permissions

ewollesen:
With the addition of full cluster lifecycle, the eksa-controller needs to be able to install / remove packages installations in the workload clusters it creates.

*Issue #, if available:*

*Description of changes:*


ewollesen:
Doh, this was totally the wrong approach, as we discussed in standup, there are better ways.

PR #868, opened 2023-03-11
ns create

ewollesen:
This is created on all clusters. The namespace will store Helm release resources, even for workload clusters.

The namespace is created when a bundle is activated.

The namespace shouldn't be created via Helm chart because that would break full cluster lifecycle installations as they run helm with the permissions of the eksa-controller-manager service account, which doesn't have permission to create namespaces.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/authenticator/target_cluster_client.go
+++ b/pkg/authenticator/target_cluster_client.go
@@ -5,9 +5,11 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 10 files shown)


pkg/authenticator/target_cluster_client.go
```diff
@@ -170,24 +176,46 @@ func (tcc *targetClusterClient) CreateClusterNamespace(ctx context.Context, clus
...
+
+	restConfig, err := tcc.ToRESTConfig()
+	if err != nil {
+		tcc.logger.V(6).Error(err, "creating rest config", "namespace", namespace)
+		return false
 	}
 
-	ns.Name = name
-	err = k8sClient.Create(ctx, ns)
+	k8sClient, err := client.New(restConfig, client.Options{})
```
lewisdiamond, line 205: A client is created in the `CreateClusterNamespace`, here and one seems to be part of `tcc` as well.

ewollesen: I was minimizing the amount of changes needed to the target cluster client. A follow up to clean that up would be good.

PR #879, opened 2023-03-16
Credential package in package controller chart

junshun:
Moving credential provider to be installed as a package via the helm chart instead of the cli.

* Update credential configurator to be backwards compatible with previous secret
* Adding defaults to credential-provider chart so it will work without having to pass to many values via the package
* Added a Package to controller chart - wrapped by cronjob.suspend in the event the cronjob is preferred
* Disabled default cronjob

After discussion it felt better to have the package controller handle the pushing secrets from management to workload clusters. 


charts/eks-anywhere-packages/templates/awssecret.yaml
```diff
@@ -1,4 +1,3 @@
-{{- if not .Values.workloadOnly }}
```
lewisdiamond, line 1: Keep the -not workload only- for now. You'd have to change the namespace otherwise. We'll only support 1 set of creds for now.

credentialproviderpackage/pkg/awscred/awscred.go
```diff
@@ -0,0 +1,55 @@
...
+		return "", err
+	}
+	secretAccessKey := strings.Trim(string(secretAccessKeyByte), "'")
+	regionByte, err := ioutil.ReadFile(regionPath)
+	if err != nil {
+		return "", err
+	}
+	region := strings.Trim(string(regionByte), "'")
+
+	awsConfig := fmt.Sprintf(
```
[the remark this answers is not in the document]
junshun: The function returns a string to compare so I dont think go template is needed here

charts/eks-anywhere-packages/templates/credentialpackage.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.cronjob.suspend }}
```
lewisdiamond, line 1: What are the cases for this to be false?

credentialproviderpackage/charts/credential-provider-package/values.yaml
```diff
@@ -12,11 +12,11 @@ image:
   pullPolicy: IfNotPresent
 
 # application values
-application:
-  secretName: aws-cred
+credentialConfiguration:
```
lewisdiamond, line 15: This should be just credentials and be an array.

charts/eks-anywhere-packages/templates/credentialpackage.yaml
junshun: They want to use the cronjob instead of credential package (for example not wanting to give node level access to registries)

lewisdiamond: We should make that configuration name more obvious then. `UseKubeletCredentials` / `UseImagePullSecret`

credentialproviderpackage/charts/credential-provider-package/templates/daemonset.yaml
```diff
@@ -50,15 +50,15 @@ spec:
...
-              value: '{{ join "," .Values.application.matchImages }}'
+              value: '{{ join "," (index .Values.credential 0).matchImages }}'
             - name: DEFAULT_CACHE_DURATION
-              value: {{.Values.application.defaultCacheDuration}}
+              value: {{(index .Values.credential 0).defaultCacheDuration}}
       volumes:
         - name: aws-creds
           secret:
-            secretName: {{.Values.application.secretName}}
+            secretName: {{(index .Values.credential 0).secretName}}
```
lewisdiamond, line 61: Add a comment mentioning we only support 1 secret for now and it's the aws-secret.

pkg/authenticator/target_cluster_client.go
```diff
@@ -219,3 +222,45 @@ func (tcc *targetClusterClient) CheckNamespace(ctx context.Context, namespace st
...
+		return fmt.Errorf("client is not initialized")
+	}
+
+	restConfig, err := tcc.ToRESTConfig()
+	if err != nil {
+		tcc.logger.V(6).Error(err, "creating rest config")
+		return err
+	}
+
+	k8sClient, err := client.New(restConfig, client.Options{})
```
lewisdiamond, line 241: We really need to fix that targetClusterClient so it handles creating the correct client(s). Out of scope here.

pkg/authenticator/target_cluster_client.go
```diff
@@ -219,3 +222,45 @@ func (tcc *targetClusterClient) CheckNamespace(ctx context.Context, namespace st
 
 	return true
 }
+
+// CreateSecret on the workload cluster
+//
+// It must only be called with an initialized target cluster client.
+func (tcc *targetClusterClient) CreateSecret(ctx context.Context, secret *corev1.Secret) error {
```
lewisdiamond, line 229: Change this function to `CreateOrUpdateSecret`, or maybe `ApplySecret` is better. Make it so that it updates the secret if it already exists, do not error in that case.

charts/eks-anywhere-packages/templates/credentialpackage.yaml
junshun: Will move this to seperate issue as this would require a CLI change to be able to activate the previous cronjob

credentialproviderpackage/pkg/awscred/awscred.go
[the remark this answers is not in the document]
junshun: changed to multiline sprintf

lewisdiamond (approved):

PR #882, opened 2023-03-18
Adding skip-webhook-validation before checking bundle is active. 

junshun:
Fixing situations where aws-secret doesn't exist and causes controller to crash

### DIFF
--- a/charts/eks-anywhere-packages/templates/credentialpackage.yaml
+++ b/charts/eks-anywhere-packages/templates/credentialpackage.yaml
@@ -6,6 +6,7 @@ metadata:
   namespace: eksa-packages-{{.Values.clusterName}}
   annotations:
     "helm.sh/resource-policy": keep
+    "skip-webhook-validation": "true"
 spec:
   packageName: credential-provider-package
   targetNamespace: eksa-packages
--- a/pkg/webhook/package_webhook.go
+++ b/pkg/webhook/package_webhook.go
@@ -58,6 +58,10 @@ func (v *packageValidator) Handle(ctx context.Context, request admission.Request
 			fmt.Errorf("decoding request: %w", err))
 	}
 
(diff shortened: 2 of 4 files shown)


charts/eks-anywhere-packages/templates/credentialpackage.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   namespace: eksa-packages-{{.Values.clusterName}}
   annotations:
     "helm.sh/resource-policy": keep
+    "skip-webhook-validation": "true"
```
lewisdiamond, line 9: Can you rename this to something like `anywhere.eks.aws.com/internal`?

junshun: done

ivyostosh: I didn't see an updated commit for this one

pkg/webhook/package_webhook.go
```diff
@@ -58,6 +58,10 @@ func (v *packageValidator) Handle(ctx context.Context, request admission.Request
 			fmt.Errorf("decoding request: %w", err))
 	}
 
+	if p.Annotations["skip-webhook-validation"] == "true" {
```
ivyostosh, line 61: This need to match the updated annotation

pkg/webhook/package_webhook.go
```diff
@@ -58,6 +58,10 @@ func (v *packageValidator) Handle(ctx context.Context, request admission.Request
 			fmt.Errorf("decoding request: %w", err))
 	}
 
+	if p.Annotations["anywhere.eks.aws.com/internal"] == "true" {
```
TerryHowe, line 61: How does this ever get set to false?

PR #887, opened 2023-03-20
Set cronjob suspend to false when cronjob is rendered.

junshun:
The rendering of the cronjob is still triggured by looking at this value vial the template function lookup-credential.method. This just fixes the case where we need to turn on the cronjob because the credential-package is not compatiable.


vincentni (approved):
/lgtm

PR #888, opened 2023-03-21
Change secret to not be reapply on management cluster. Update BR versions

junshun:
BR v1.13 was released and does not include the fix needed for credential package to work so logic was updated.

Removing the need to copy annotations for secret over as secret is no longer reapplied on management cluster.


TerryHowe (approved):
/lgtm

PR #890, opened 2023-03-22
Add support for rendering just a package

lewisdiamond:
*Description of changes:* Add support for only rendering a package resource as a temporary workaround until that's generated by the eksa controller.

/hold until I review it tomorrow

### DIFF
--- a/charts/eks-anywhere-packages/templates/_helpers.tpl
+++ b/charts/eks-anywhere-packages/templates/_helpers.tpl
@@ -147,3 +147,18 @@ Function to figure out if to install cronjob, credential-package, or none
         {{- printf "none" -}}
     {{- end -}}
 {{- end -}}
+
+
+{{/*
+RenderType returns the type of output we need to render, either controller, workload or package.
+*/}}
+{{- define "eks-anywhere-packages.rendertype" -}}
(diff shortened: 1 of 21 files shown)


charts/eks-anywhere-packages/templates/_helpers.tpl
```diff
@@ -147,3 +147,18 @@ Function to figure out if to install cronjob, credential-package, or none
...
+RenderType returns the type of output we need to render, either controller, workload or package.
+*/}}
+{{- define "eks-anywhere-packages.rendertype" -}}
+    {{- if and (not .Values.workloadOnly) (not .Values.workloadPackageOnly) -}}
+        {{- printf "controller" -}}
+    {{- else if .Values.workloadPackageOnly -}}
+        {{- printf "package" -}}
+    {{- else -}}
+        {{- printf "workload" -}}
+    {{- end -}}
```
ewollesen, line 162:
```suggestion
    {{- if .Values.workloadPackageOnly -}}
        {{- printf "package" -}}
    {{- else if .Values.workloadOnly -}}
        {{- printf "workload" -}}
    {{- else -}}
        {{- printf "controller" -}}
    {{- end -}}
```
I feel like this is easier to understand, and is functionally equivalent.

PR #891, opened 2023-03-22
Control Plane Toleration and Bottlerocket version check

junshun:
* Added CP toleration as default for credential package
* Add version check to configurator binary to prevent node breakage
* Defaulting to cronjob for BR regardless of k8s version for BR < 1.13 by default

### DIFF
--- a/credentialproviderpackage/go.mod
+++ b/credentialproviderpackage/go.mod
@@ -3,11 +3,11 @@ module github.com/aws/eks-anywhere-packages/credentialproviderpackage
 go 1.19
 
 require (
-	github.com/Masterminds/semver v1.5.0
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/google/go-cmp v0.5.9
 	github.com/mitchellh/go-ps v1.0.0
 	github.com/stretchr/testify v1.8.0
+	golang.org/x/mod v0.3.0
 	k8s.io/apimachinery v0.26.0
 	sigs.k8s.io/yaml v1.3.0
 )
--- a/credentialproviderpackage/go.mod
+++ b/credentialproviderpackage/go.mod
@@ -3,6 +3,7 @@ module github.com/aws/eks-anywhere-packages/credentialproviderpackage
 go 1.19
 
 require (
+	github.com/Masterminds/semver v1.5.0
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/google/go-cmp v0.5.9
 	github.com/mitchellh/go-ps v1.0.0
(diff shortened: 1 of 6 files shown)


credentialproviderpackage/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/aws/eks-anywhere-packages/credentialproviderpackage
 go 1.19
 
 require (
+	github.com/Masterminds/semver v1.5.0
```
lewisdiamond, line 6: Can you use `"golang.org/x/mod/semver"`
We use it here: https://github.com/aws/eks-anywhere-packages/blob/main/pkg/bundle/manager.go#L10

credentialproviderpackage/pkg/configurator/bottlerocket/bottlerocket.go
```diff
@@ -56,19 +69,26 @@ func NewBottleRocketConfigurator(socketPath string) (*bottleRocket, error) {
 	if socket.Mode().Type() != fs.ModeSocket {
 		return nil, fmt.Errorf("Unexpected type %s expected socket\n", socket.Mode().Type())
 	}
-	return &bottleRocket{
+
+	br := &bottleRocket{
+		baseURL: "http://localhost/",
```
lewisdiamond, line 74: What makes this necessary now?

credentialproviderpackage/pkg/configurator/bottlerocket/bottlerocket.go
```diff
@@ -179,3 +199,69 @@ func createCredentialProviderPayload(config constants.CredentialProviderConfigOp
 	}
 	return payload, nil
 }
+
+func (b *bottleRocket) verifySupportedBRVersion() error {
```
lewisdiamond, line 203: I'd make it `isSupportedBRVersion() (bool, error)` true/false means supported or not and an error means something wrong happened.

credentialproviderpackage/pkg/configurator/bottlerocket/bottlerocket_test.go
```diff
@@ -340,11 +341,111 @@ func Test_bottleRocket_Initialize(t *testing.T) {
...
+			wantErr:    true,
+		},
+		{
+			name:       "bad response from server",
+			fields:     fields{client: http.Client{}},
+			brVersion:  "1.13.1",
+			brVariant:  "vmware-k8s-1.25",
+			statusCode: http.StatusNotFound,
+			wantErr:    true,
+		},
```
lewisdiamond, line 414: Can you add a test case where BR returns new fields, missing fields, etc?

PR #910, opened 2023-03-30
Commenting out packages for artifact-move

jonahjon:
*Issue #, if available:*

*Description of changes:*

For the first release we don't want to move any other helm charts than the EKS-A release bundle version (Which already exists) for packages into the prod ECR repo as it's confusing to a user. 

We will un-comment this once we have newer versions we want to update later on.

### DIFF
--- a/generatebundlefile/data/prod_artifact_move.yaml
+++ b/generatebundlefile/data/prod_artifact_move.yaml
@@ -4,12 +4,6 @@ kubernetesVersion: "1.22"
 packages:
   - org: aws
     projects:
-      # - name: eks-anywhere-packages
-      #   copyimages: true
-      #   repository: eks-anywhere-packages
-      #   registry: public.ecr.aws/l0g8r8j6
-      #   versions:
-      #     - name: 0.0.0-latest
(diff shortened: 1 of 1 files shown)


generatebundlefile/data/prod_artifact_move.yaml
```diff
@@ -4,12 +4,12 @@ kubernetesVersion: "1.22"
 packages:
   - org: aws
     projects:
-      - name: eks-anywhere-packages
-        copyimages: true
-        repository: eks-anywhere-packages
-        registry: public.ecr.aws/l0g8r8j6
-        versions:
-          - name: 0.3.9-eks-a-33
+      # - name: eks-anywhere-packages
```
a-cool-train, line 7: We should probably remove it if we won't be using it.

jonahjon: We want to add this back in during the next update so I just kept it there to help people, but can remove it too.

jonahjon: removed

PR #922, opened 2023-04-14
Template credential-provider-config api version

junshun:
Have credential-provider-config api version be dynamically generated based on kubernetes version.
Kubernetes added v1 support for kubelet.config.k8s.io and credentialprovider.kubelet.k8s.io in v1.26 and will be removing v1alpha1 in future versions. Adding support now before v1.27

Currently this is setup to bump the api version one version after the release supported. For example even though v1 was added to v1.26 we will still use v1beta1. This is because there are cases during an upgrade for example of a rolling upgrade of v1.25 -> v1.26 cluster that the node the helm chart installs on might be v1.26 and try to run the deamonset on the 1.25 nodes with an unsupported api. Since we don't allow a two version jump of k8s version at once this should be fine.

### DIFF
--- a/credentialproviderpackage/pkg/configurator/linux/linux_test.go
+++ b/credentialproviderpackage/pkg/configurator/linux/linux_test.go
@@ -163,11 +163,8 @@ func Test_linuxOS_updateKubeletArguments(t *testing.T) {
 				basePath:      tt.fields.basePath,
 				config:        tt.fields.config,
 			}
-			err := os.Setenv("K8S_VERSION", tt.k8sVersion)
-			if err != nil {
-				t.Errorf("updateKubeletArguments failed to set K8S_VERSION")
-			}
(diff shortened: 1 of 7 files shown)


credentialproviderpackage/pkg/configurator/linux/linux_test.go
```diff
@@ -108,6 +163,11 @@ func Test_linuxOS_updateKubeletArguments(t *testing.T) {
 				basePath:      tt.fields.basePath,
 				config:        tt.fields.config,
 			}
+			err := os.Setenv("K8S_VERSION", tt.k8sVersion)
```
TerryHowe, line 166: Why not t.Setenv here?

junshun: done

TerryHowe (approved):
/lgtm

PR #930, opened 2023-05-01
Update to go 1.20

TerryHowe:


scripts/common.sh
```diff
@@ -84,6 +84,8 @@ function build::common::upload_artifacts() {
 }
 
 function build::gather_licenses() {
+  # force 1.16 since thats the version used to install go-licenses in builder-base
+  build::common::use_go_version 1.18
```
junshun, line 88: Seems to be mismatch here with comment in code. Should it be either 1.16 or 1.18?

scripts/common.sh
```diff
@@ -84,6 +84,8 @@ function build::common::upload_artifacts() {
 }
 
 function build::gather_licenses() {
+  # force 1.16 since thats the version used to install go-licenses in builder-base
```
TerryHowe, line 87:
```suggestion
  # force 1.18 since thats the version used to install go-licenses in builder-base
```

PR #931, opened 2023-05-02
Update helm version

TerryHowe:
Updating helm required the updating of several other things which in turn caused more updating.


TerryHowe:
/hold

controller-gen broken, but want to get this out there

ivyostosh:
Checked out locally and it ran successfully. However, I didn't manage to test this in a new EKS-A cluster.

PR #955, opened 2023-06-27
Add awsconfig value

d8660091:
*Issue #, if available:* the users should be able to auth through aws config

*Description of changes:* Add awsconfig field to values.yaml

### DIFF
--- a/charts/eks-anywhere-packages/templates/awssecret.yaml
+++ b/charts/eks-anywhere-packages/templates/awssecret.yaml
@@ -13,6 +13,9 @@ data:
   AWS_ACCESS_KEY_ID: "{{ .id }}"
   AWS_SECRET_ACCESS_KEY: "{{ .secret }}"
   REGION: "{{ .region }}"
+  {{- if .config }}
+  config: "{{ .config }}"
+  {{- end }}
   {{- end }}
 type: Opaque
 {{- end }}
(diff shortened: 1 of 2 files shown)


charts/eks-anywhere-packages/templates/credentialpackage.yaml
```diff
@@ -26,7 +26,7 @@ spec:
       - matchImages:
         - {{.Values.defaultImageRegistry}}
         profile: "default"
-        secretName: aws-secret
+        secretName: {{ ternary "aws-secret" "aws-config" (empty .Values.awsConfig) }}
```
TerryHowe, line 29: Why are we adding a new secret instead of just modifying aws-secret? What other methods of authn is this adding?

d8660091: Good point. I created a new secret by following this doc https://anywhere.eks.amazonaws.com/docs/packages/credential-provider-package/iam_roles_anywhere/. But after checking it in details, I think it's possible to reuse aws-secret.

> What other methods of authn is this adding?
By making this aws-secret general, the users will be able to use "role-assume", "aws roles anywhere" etc to do authn, reference: https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html.

TerryHowe (approved):
/lgtm

PR #962, opened 2023-07-07
Adjust eks-anywhere-packages tags to sha256 format

d8660091:
*Issue #, if available:* In airgapped environment, registry-mirror needs to host curated packages images. Those images' urls can only be fetched from packageBundle. But packageBundle doesn't store tags information. So registry-mirror will not have tag info as well.

*Description of changes:* eks-anywhere-package helm chart and its dependencies switch to sha256 tag so it can pull images from registry-mirror

### DIFF
--- a/charts/eks-anywhere-packages-migrations/templates/_helpers.tpl
+++ b/charts/eks-anywhere-packages-migrations/templates/_helpers.tpl
@@ -2,5 +2,9 @@
 Create image name
 */}}
 {{- define "template.image" -}}
-{{- printf "/%s@%s" .repository .digest -}}
+{{- if eq (substr 0 7 .tag) "sha256:" -}}
+{{- printf "/%s@%s" .repository .tag -}}
+{{- else -}}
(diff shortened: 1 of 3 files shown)


charts/eks-anywhere-packages-migrations/templates/_helpers.tpl
```diff
@@ -2,5 +2,9 @@
 Create image name
 */}}
 {{- define "template.image" -}}
```
TerryHowe, line 4: If the package bundle doesn't have tags, why are we using it here? Should this use the digest first and tag second?

d8660091: It's changed back mainly to keep align with the existing two definitions: https://github.com/search?q=repo%3Aaws%2Feks-anywhere-packages+define+%22template.image%22&type=code

TerryHowe (approved):
/lgtm

PR #966, opened 2023-08-02
Use real tag in eks-anywhere-packages

d8660091:
*Issue #, if available:* we need to use real tag inside eks-anywhere-packages because some hack during release process depends on it.

*Description of changes:*


TerryHowe (approved):
/lgtm

PR #978, opened 2023-09-11
Add ecr-cred-injector

d8660091:
*Issue #, if available:* 

*Description of changes:* This change will allow packages controller to be able to pull charts from private ECR. It injects the credential(username:password) to the docker config file of the package controller container.

Manually tested that the injector works for regional ECR


pkg/registry/ecr_cred_injector.go
```diff
@@ -0,0 +1,255 @@
...
+		return err
+	}
+
+	registry, err := a.GetECR(ctx)
+	if err != nil {
+		return err
+	}
+
+	if !IsECRRegistry(registry) {
+		a.log.Info("defaultRegistry is not ECR registry, skip injecting credential to docker config")
```
jonathanmeier5, line 89: are we missing a return here?

pkg/registry/ecr_cred_injector.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &secret, nil
+}
+
+// InjectCredentail update field "config.json" in the secret, which is used by packages controller's oras and helm
+func (a *ECRCredInjector) InjectCredential(ctx context.Context, secret v1.Secret, registry string, cred auth.Credential) error {
+	d := secret.Data["config.json"]
```
jonathanmeier5, line 129: do we need to verify this key exists?

charts/eks-anywhere-packages/templates/deployment.yaml
```diff
@@ -146,6 +148,10 @@ spec:
           defaultMode: 420
           secretName: ecr-token
           optional: true
+      - name: aws-secret
```
jonathanmeier5, line 151: what do you think about `aws-config-secret`? or is `aws-secret` the convention elsewhere?

checking my understanding: we can't use the existing ecr-token secret here because we don't have control over its renewal lifecycle?

d8660091: "aws-secret" contains a field called "config", calling it "aws-config-secret" may imply that the volume is the "config" of "aws-secret".

ecr-token stores "username:password", it's not refreshed anymore perhaps because we have migrated to use credential-provider.

pkg/registry/ecr_cred_injector.go
```diff
@@ -0,0 +1,255 @@
...
+		Name: authenticator.MirrorCredName,
+	}, &secret)
+	if err != nil {
+		return nil, err
+	}
+
+	return &secret, nil
+}
+
+// InjectCredentail update field "config.json" in the secret, which is used by packages controller's oras and helm
```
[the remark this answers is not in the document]
d8660091: good caught

pkg/registry/ecr_cred_injector.go
d8660091: It should be ok, if this dockerconfig file doesn't exist, it would need be created so we can inject the ECR credential

ecrtokenrefresher/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY pkg/ pkg/
+COPY cmd/ cmd/
+
+# Build
+RUN GOPROXY=direct CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o ecr-token-refresh ./cmd/ecr-token-refresher
+
+FROM gcr.io/distroless/static:debug-nonroot
+WORKDIR /
+COPY --from=builder /workspace/ecr-token-refresh .
+USER 65532:65532
```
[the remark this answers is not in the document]
d8660091: good idea

jonathanmeier5 (approved):
/lgtm

PR #1001, opened 2023-10-17
Update staging to use properly versioned tags

tatlat:
*Issue #, if available:*

*Description of changes:*


generatebundlefile/data/bundles_staging/1-23.yaml
```diff
@@ -9,7 +9,7 @@ packages:
         repository: eks-anywhere-packages
         registry: public.ecr.aws/w9m0f3l5
         versions:
-          - name: 0.0.0-8862036270224f2a6b8d6ecd455b6b1fa1084619
+          - name: v0.3.12-f567db688066bcbb3516e4bc14662796ec920906
```
d8660091, line 12: we need a tag without "v" prefix, helm chart tag doesn't start with "v".

PR #1005, opened 2023-10-25
Support private ECR-only bundle generation and promotion

abhay-krishna:
Support private ECR-only bundle generation and promotion.

### DIFF
--- a/generatebundlefile/hack/docker-ecr-config.json
+++ b/generatebundlefile/hack/docker-ecr-config.json
@@ -7,6 +7,7 @@
         "382577505035.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
         "646717423341.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
         "783794618700.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
+        "067575901363.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
         "public.ecr.aws": "ecr-login"
     }
 }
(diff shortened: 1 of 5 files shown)


generatebundlefile/hack/docker-ecr-config.json
```diff
@@ -7,6 +7,7 @@
         "382577505035.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
         "646717423341.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
         "783794618700.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
+        "067575901363.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
```
jonathanmeier5, line 10: do we need all the other registries added here too?

jonathanmeier5:
what does NewBuildMode do?

generatebundlefile/hack/docker-ecr-config.json
abhay-krishna: Since we're building only to this account, this should suffice. The replication happens outside of this Go code.

abhay-krishna:
> what does NewBuildMode do?

Currently we're running builds from the other CDK's build pipelines too, and both share the same repo. Currently the process involves public ECR operations which wouldn't work for the new packages pipelines since they're pushing to regional ECR and not public ECR. So on the packages CDK build projects, we're setting the env var and checking for it here and that helps us perform private ECR-only client creation and operations.

jonathanmeier5 (approved):

PR #1010, opened 2023-10-26
Add packages input files for private ECR bundle generation

abhay-krishna:
Add packages input files for private ECR bundle generation.


d8660091:
Would it be better to call the yaml file "regional" instead of "new"? And will each codebuild run populate just 1 region?

abhay-krishna:
> Would it be better to call the yaml file "regional" instead of "new"? And will each codebuild run populate just 1 region?

Yeah I like that naming. Will change it. And yes it's building in one region only and being replicated to other regions.

d8660091 (commented):
/LGTM

PR #1021, opened 2023-11-01
Add input files for regional staging package bundle generation

abhay-krishna:
* Add input files for regional staging package bundle generation 
* Promote `eks-anywhere-packages` (and related artifacts), `ecr-token-refresher`, `credential-provider-package` to public ECR.

### DIFF
--- a/generatebundlefile/promote.go
+++ b/generatebundlefile/promote.go
@@ -49,17 +49,16 @@ func GetSDKClients(regionalBuildMode bool) (*SDKClients, error) {
 		clients.stsClientRelease = clients.stsClient
 		clients.ecrClientRelease = clients.ecrClient
 	}
-	if !regionalBuildMode {
-		// ECR Public Connection with us-east-1 region
-		conf, err = config.LoadDefaultConfig(context.TODO(), config.WithRegion(ecrPublicRegion))
-		if err != nil {
-			return nil, fmt.Errorf("loading default AWS config: %w", err)
-		}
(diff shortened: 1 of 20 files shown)


generatebundlefile/promote.go
```diff
@@ -297,13 +302,30 @@ func (c *SDKClients) moveImage(image Image, authFile string) error {
...
 	if err != nil {
-		return fmt.Errorf("unable to copy image digest from source to destination repo %w", err)
+		return fmt.Errorf("unable to copy image digest from source to destination repo: %w", err)
+	}
+
+	if regionalBuildMode {
+		destinationPrefix = fmt.Sprintf("docker://%s/%s", c.ecrPublicClient.SourceRegistry, image.Repository)
+		BundleLog.Info(fmt.Sprintf("Promoting images to secondary ECR destination: %s/%s", c.ecrPublicClient.SourceRegistry, image.Repository))
+
+		err := copyImage(BundleLog, authFile, sourcePrefix+":"+image.Tag, destinationPrefix+":"+image.Tag)
```
d8660091, line 318: Q: how is it guaranteed that only the dependencies of curated-packages controller moved to the public registry? Looks like copyImage doesn't do this check.

abhay-krishna: We do that check [here](https://github.com/aws/eks-anywhere-packages/blob/main/generatebundlefile/hack/promote.sh#L37)

d8660091 (approved):

PR #1022, opened 2023-11-03
Add regional staging artifact move file

abhay-krishna:
Add regional staging artifact move file and update some tags.


generatebundlefile/hack/common.sh
```diff
@@ -29,7 +29,14 @@ function awsAuth () {
         region="--region=us-east-1"
     fi
 
-    aws "$awsCmd" "$region" "--profile=${PROFILE:-}" get-login-password
```
abhay-krishna, line 32: Moving the oras login inside the function since we don't use the `awsAuth` function anywhere else.

generatebundlefile/hack/release_staging.sh
```diff
@@ -21,8 +21,17 @@ set -o pipefail
...
+BASE_DIRECTORY=$(git rev-parse --show-toplevel)
+
+. "${BASE_DIRECTORY}/generatebundlefile/hack/common.sh"
+ORAS_BIN=${BASE_DIRECTORY}/bin/oras
 
-cat << EOF > configfile
+make build
+chmod +x ${BASE_DIRECTORY}/generatebundlefile/bin/generatebundlefile
+
+regional_build_mode=${REGIONAL_BUILD_MODE:-}
```
abhay-krishna, line 32: Only in the non-regional mode (old flow), we push packages to the old packages account, otherwise we skip it and just push to the staging account (`public.ecr.aws/w9m0f3l5`)

generatebundlefile/main.go
```diff
@@ -138,11 +142,31 @@ func cmdPromote(opts *Options) error {
 		},
 	}
 
+	Profile := "default"
+	val, ok := os.LookupEnv("AWS_PROFILE")
+	if ok {
+		Profile = val
+	}
+	if opts.regionalBuildMode && Profile != "default" {
```
abhay-krishna, line 150: If we are running the promote in regional mode and profile is staging/prod, we need a public ECR release clients (corresponding to staging or prod accounts), else the default dev clients will suffice.

generatebundlefile/promote.go
```diff
@@ -112,16 +114,31 @@ func (c *SDKClients) PromoteHelmChart(repository, authFile, tag string, copyImag
 	// If we are not moving artifacts to the Private ECR Packages artifact account, get information from public ECR instead.
 	// The first scenario runs for flags --private-profile and --promote.
 	// The 2nd scenario runs for flags and --public-profile.
-	if c.ecrClientRelease != nil || c.ecrClient != nil {
-		name, version, sha, err = c.getNameAndVersion(repository, tag, c.stsClient.AccountID)
-		if err != nil {
-			return fmt.Errorf("Error getting name and version from helmchart %s", err)
+	if regionalBuildMode {
```
abhay-krishna, line 117: If running in regional build mode, the helm chart destination is the artifacts account public ECR hence we get the name and version from there, else we get it from private ECR.

PR #1045, opened 2024-01-08
Fix credential provider when deployed to Ubuntu 22.04

chrisdoherty4:
Closes #1814


credentialproviderpackage/charts/credential-provider-package/templates/daemonset.yaml
```diff
@@ -67,11 +67,6 @@ spec:
         - name: socket
           hostPath:
             path: /run/api.sock
-        {{- else if eq $os "default"}}
```
d8660091, line 70: Does Amazon Linux work with "/etc/sysconfig/kubelet"?

chrisdoherty4: I'm sorry, I see what you're saying. I'll revert this change.

PR #1095, opened 2024-04-04
Provide admission webhook decoder when instantiating CRD validators

abhay-krishna:
Recently we bumped controller-runtime (CR) from v0.14.4 to v0.16.5. In CR v0.15.0, they added a [change](https://github.com/kubernetes-sigs/controller-runtime/commit/927c09e54313d6476b5780b7f029f381d777c149) which removes logic for the injection of decoders. Hence we need to explicitly initialize the decoders when registering the validating webhook handlers for the packages CRDs.

### DIFF
--- a/pkg/packages/manager_test.go
+++ b/pkg/packages/manager_test.go
@@ -31,14 +31,14 @@ models:
   cj2a:
     year: "45"
 `
-)
-const newConfiguration = `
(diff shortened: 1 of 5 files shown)


abhinavmpandey08 (approved):
/woof
/lgtm
/approve

PR #1105, opened 2024-04-18
Updating registry mirror CA cert in PBC controller reconcile

pokearu:
*Issue https://github.com/aws/eks-anywhere-internal/issues/2269:*

*Description of changes:*
Packages controller pod has a mounted secret `registry-mirror-cred`. This secret creates the CA cert on the pod with the name `<cluster-name>_ca.crt`.

When dealing with multiple clusters; specifically when workload clusters are created from EKS-A. The packages controller pod fails with an 
```
tls: failed to verify certificate: x509: certificate signed by unknown authority
```
As the new cluster's CA cert is not updated on the mounted secret.

These changes add a step in the PBC controller workflow that checks the mounted secret and updates the registry CA cert if the field is missing in the data.


abhinavmpandey08 (approved):
/approve
/lgtm

PR #1126, opened 2024-06-21
Fix metrics server image tags for all Kubernetes versions

rahulbabu95:
*Description of changes:*
The metric server images in build account for Kubernetes version 1.27 and above are tagged with suffix `-latest-helm`. Add the suffix to fix build failures on packages staging.

### DIFF
--- a/generatebundlefile/data/staging_artifact_move.yaml
+++ b/generatebundlefile/data/staging_artifact_move.yaml
@@ -88,10 +88,10 @@ packages:
         registry: public.ecr.aws/l0g8r8j6
         versions:
           - name: 0.7.1-eks-1-26-38-fe14b98f98132123f478eaf6e65bba4a29bf488b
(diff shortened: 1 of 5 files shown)


generatebundlefile/data/staging_artifact_move.yaml
```diff
@@ -88,10 +88,10 @@ packages:
         registry: public.ecr.aws/l0g8r8j6
         versions:
           - name: 0.7.1-eks-1-26-38-fe14b98f98132123f478eaf6e65bba4a29bf488b
```
sp1999, line 90: Did you miss the suffix here?

rahulbabu95: no it's only for images from 1.27 and [above](https://tiny.amazon.com/1du29girw/IsenLink). The promotion logic is weird and depends on the time it runs and which image is already having the `latest` tag at that time. It certainly needs to be revisited but to your question 1-26 doesn't have that tag on our public ECR.

PR #1128, opened 2024-06-21
Filter out proper image tags based on tag format

abhay-krishna:
We want the promoted image tags to be of the format `X.Y.Z-<commit hash>`, for example, `0.4.3-451cb0d2fd416176540b65be9e46b4cfc0db99f0` but sometimes if the same digest has multiple tags, it could pick up an image which does not follow the structure, such as `0.4.3-latest-helm`, as it's just getting the first element from the list without checking its format.

### DIFF
--- a/generatebundlefile/ecr_helper.go
+++ b/generatebundlefile/ecr_helper.go
@@ -186,7 +186,16 @@ func getLatestHelmTagandSha(details []ImageDetailsBothECR) (string, string, erro
 	if reflect.DeepEqual(latest, ImageDetailsBothECR{}) {
 		return "", "", fmt.Errorf("error no images found")
 	}
-	return latest.ImageTags[0], *latest.ImageDigest, nil
+
+	var imageTag string
+	imageTagRegex := regexp.MustCompile(`^\d+\.\d+\.\d+.*[0-9a-f]{40}$`)
+	for _, tag := range latest.ImageTags {
+		if imageTagRegex.MatchString(tag) {
+			imageTag = tag
+			break
+		}
+	}
+	return imageTag, *latest.ImageDigest, nil
 }
 
 // getLatestImageSha Iterates list of ECR Public Helm Charts, to find latest pushed image and return tag/sha  of the latest pushed image


generatebundlefile/ecr_helper.go
```diff
@@ -186,7 +186,16 @@ func getLatestHelmTagandSha(details []ImageDetailsBothECR) (string, string, erro
 	if reflect.DeepEqual(latest, ImageDetailsBothECR{}) {
 		return "", "", fmt.Errorf("error no images found")
 	}
-	return latest.ImageTags[0], *latest.ImageDigest, nil
+
+	var imageTag string
+	imageTagRegex := regexp.MustCompile(`^\d+\.\d+\.\d+-[0-9a-f]{40}$`)
```
rahulbabu95, line 191: Does this handle eks-d tag that follows a version. Not sure what this regex is trying to capture but tried running it online against a sample tag we might have for metrics server: `9.37.0-1.27-d18f9e2def8e8e24b3ce0e0c543cc4d1d7d6512e` and it failed.

generatebundlefile/ecr_helper.go
```diff
@@ -186,7 +186,16 @@ func getLatestHelmTagandSha(details []ImageDetailsBothECR) (string, string, erro
...
+	for _, tag := range latest.ImageTags {
+		if imageTagRegex.MatchString(tag) {
```
rahulbabu95, line 193: nit: if possible it will be nice to also filter out a latest-helm tag from a commit cos i saw ran into this yesterday where the private ECR has images with tags both `0.7.1-eks-1-27-32-fe14b98f98132123f478eaf6e65bba4a29bf488b` and `0.7.1-eks-1-27-32-fe14b98f98132123f478eaf6e65bba4a29bf488b-latest-helm` this and during the promote it somehow ends up promoting the one with suffix latest-helm.

generatebundlefile/ecr_helper.go
abhay-krishna: good catch, fixed the regex

generatebundlefile/ecr_helper.go
abhay-krishna: The regex should filter out the latest helm because it will only accept tags which end at the SHA (note the `$` at the end of the regex)

PR #1142, opened 2024-07-16
Remove helm chart promotion to public ECR in staging and prod release

sp1999:
*Description of changes:*
This PR remove unnecessary helm charts promotion to public ECR registry in staging and prod release jobs as we now handle that in the ECR replication job. It also refactors the script to call the `generate` and `push` bundle functions in a single loop over kubernetes versions and update the comments to reflect what the job actually does.

### DIFF
--- a/generatebundlefile/hack/release_prod.sh
+++ b/generatebundlefile/hack/release_prod.sh
@@ -29,47 +29,19 @@ ORAS_BIN=${BASE_DIRECTORY}/bin/oras
 make build
 chmod +x ${BASE_DIRECTORY}/generatebundlefile/bin/generatebundlefile
 
-# Release Helm Chart, and bundle to Production account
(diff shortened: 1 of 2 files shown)


generatebundlefile/hack/release_prod.sh
```diff
@@ -44,33 +44,19 @@ credential_source=EcsContainer
 EOF
 
 export AWS_CONFIG_FILE=${BASE_DIRECTORY}/generatebundlefile/prodconfigfile
```
abhay-krishna, line 46: Do we still need to keep this config file? we are not assuming any role in a different account, so we can probably remove this and the codeblocks that are executed when the profile is not default.

sp1999: Updated

PR #1143, opened 2024-07-17
Tag package bundles with epoch time and date of creation

abhay-krishna:
We are changing the logic of packages bundles tagging to use the date and epoch time of the bundle creation instead of an arbitrary value like the CodeBuild build number.

### DIFF
--- a/generatebundlefile/input.go
+++ b/generatebundlefile/input.go
@@ -4,9 +4,9 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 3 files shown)


generatebundlefile/input.go
```diff
@@ -142,14 +143,13 @@ func (c *SDKClients) NewBundleFromInput(Input *Input) (api.PackageBundleSpec, st
...
-		version := strings.Split(Input.KubernetesVersion, ".")
-		name = fmt.Sprintf("v1-%s-%s", version[1], name)
-	}
+
+	currentTime := time.Now()
+	currentEpochTime := currentTime.Unix()
+	date := currentTime.Format(YYYYMMDDFormat)
+
+	version := strings.Split(Input.KubernetesVersion, ".")
+	name := fmt.Sprintf("v1-%s-%d-%s", version[1], currentEpochTime, date)
```
vivek-koppuru, line 152:
```suggestion
	name := fmt.Sprintf("v1-%s-%s-%d", version[1], date, currentEpochTime)
```

PR #1157, opened 2024-08-23
Use helm charts only from main branch when finding latest image tag

sp1999:
*Description of changes:*
Currently, the dev packages bundles are generated based on the dev bundle input files defined [here](https://github.com/aws/eks-anywhere-packages/tree/main/generatebundlefile/data/bundles_dev) for each kubernetes version. Except for the package controller helm charts, all other packages helm charts have the "latest" version tag in the input file which tells the program to use a timestamp lookup and find the most recently pushed helm chart for information, even if that helm chart doesn't have the latest tag. This causes an issue when a certain package gets built from the release branch as that tag will be used in the dev packages bundle instead of the latest tag. Even if the packages helm chart have the same version/tag on the main and release branch, it could still cause an issue when certain bug fixes from main aren't backported to the release branch in the [eks-anywhere-build-tooling](https://github.com/aws/eks-anywhere-build-tooling) repository. For example, the cert-manager tests started failing recently when the cert-manager helm chart built from the release-0.20 branch was used in the generated packages dev bundle due to a bug fix merged in [#3403](https://github.com/aws/eks-anywhere-build-tooling/pull/3403) not being backported to the release-0.20 branch. This PR updates the logic in the generatebundlefile CLI to find the most recently pushed helm chart which also contains the "latest" tag so that it doesn't use the helm charts from the release branches.

### DIFF
--- a/generatebundlefile/README.md
+++ b/generatebundlefile/README.md
@@ -80,7 +80,7 @@ packages:
 
 #### Latest Version
 
-The "latest" tag will tell the program to use a timestamp lookup, and find the most recently pushed helm chart for information, even if that helm chart doesn't have the latest tag.
+The "latest" tag will tell the program to use a timestamp lookup, and find the most recently pushed helm chart for information only if that helm chart has the latest tag.
 
 ```yaml
 name: "v1-22-1001"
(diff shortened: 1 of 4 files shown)


abhay-krishna (approved):
/approve
/lgtm

PR #1175, opened 2024-11-07
Add update_bundle script to get latest version and update input bundle for staging and prod

jhaanvi5:
### Description of changes:
This PR introduces a new `update_bundles.sh` script that automates the process of updating input bundle files for both staging and production bundles for packages release. The script fetches the latest versions of packages from ECR and updates the corresponding YAML files for Kubernetes versions. This addition will streamline our bundle update workflow during packages release and reduce manual errors.

#### Testing
Tested script for both staging and prod bundles.
`sh update_bundles.sh -e staging`

### DIFF
--- a/generatebundlefile/hack/update_bundles.sh
+++ b/generatebundlefile/hack/update_bundles.sh
@@ -1,0 +1,161 @@
+#!/bin/bash
+
+set -e
+set -o pipefail
+
+BASE_DIRECTORY=$(git rev-parse --show-toplevel)
+# Function to display usage information
+function usage() {
+  echo "Usage: $0 -e <environment>"
+  echo "  -e    Specify the environment (staging or prod)"
+  exit 1
+}
+
+# Function to check if required tools are installed
+function check_requirements() {
+  local required_tools=("aws" "jq" "yq")
+  for tool in "${required_tools[@]}"; do
+    if ! command -v "$tool" &>/dev/null; then
+      echo "Error: $tool is not installed or not in PATH" >&2
+      exit 1
+    fi
+  done
+}
+
+# Function to get the latest version from ECR
+function get_latest_version() {
+  local repo=$1
+  local k8s_version=$2
+
+  local json_output=$(aws ecr describe-images \
+    --repository-name "$repo" \
+    --output json)
+
+  # For metrics-server and cluster autoscaler k8s version is embedded in image tag
+  # Cluster Autosclaer has version format 1.XX and metrics-server has format 1-XX
+  # For cluster autoscaler filter tag based on k8s version (1.28,1.29,etc)
+  if [ "$repo" = "cluster-autoscaler/charts/cluster-autoscaler" ]; then
+    kube_version="1.$k8s_version"
+  # For metrics-server filter tag based on k8s version (1-28,1-29,etc)
+  elif [ "$repo" = "metrics-server/charts/metrics-server" ]; then
+    kube_version="1-$k8s_version"
+  fi
+
+  if [ -n "${kube_version:-}" ]; then
+    output=$(echo $json_output | jq -r --arg kv "$kube_version" '
+    .imageDetails
+    | map(select(.imageTags | length > 0))
+    | map(select(.imageTags | map(test($kv)) | any))
+    | sort_by(.imagePushedAt)
+    | last
+    | .imageTags')
+  else
+    output=$(echo $json_output | jq -r '
+    .imageDetails
+    | sort_by(.imagePushedAt)
+    | last(.[]).imageTags')
+  fi
+
+  # Convert tags to array
+  local versions=()
+  for value in $(echo "$output" | jq -r '.[]'); do
+    versions+=("$value")
+  done
+
+  # Get latest version and remove tags with helm
+  local latest_version=$(printf "%s\n" "${versions[@]}" |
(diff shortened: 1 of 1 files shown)


generatebundlefile/hack/update_bundles.sh
```diff
@@ -0,0 +1,150 @@
...
+  if [[ "$environment" != "staging" && "$environment" != "prod" ]]; then
+    echo "Error: Invalid environment. Use 'staging' or 'prod'" >&2
+    usage
+  fi
+
+  # Check if required tools(aws,jq,yq) are installed
+  check_requirements
+
+  # Generate bundle files for Kubernetes versions 1.27 to 1.31 for given environment
+  for version in {27..31}; do
```
vivek-koppuru, line 142: We should ideally get this from a config file or from folder structure instead of hardcoded

generatebundlefile/hack/update_bundles.sh
```diff
@@ -0,0 +1,150 @@
...
+    | .imageTags')
+  # For metrics-server filter tag based on k8s version (1-28,1-29,etc)
+  elif [ "$repo" = "metrics-server/charts/metrics-server" ]; then
+    output=$(echo $json_output | jq -r --arg kv "1-$k8s_version" '
+    .imageDetails
+    | map(select(.imageTags | length > 0))
+    | map(select(.imageTags | map(test($kv)) | any))
+    | sort_by(.imagePushedAt)
+    | last
+    | .imageTags')
```
vivek-koppuru, line 51: We can optimize this to not have this repeated code right?

vivek-koppuru (commented):
Could be worth thinking about writing this in Go

generatebundlefile/hack/update_bundles.sh
```diff
@@ -0,0 +1,150 @@
...
+  for tool in "${required_tools[@]}"; do
+    if ! command -v "$tool" &>/dev/null; then
+      echo "Error: $tool is not installed or not in PATH" >&2
+      exit 1
+    fi
+  done
+}
+
+# Function to get the latest version from ECR
+function get_latest_version() {
```
vivek-koppuru, line 26: Should we cross verify this somehow instead of just looking at latest, like GIT_TAG or anything else?

jhaanvi5:
> Could be worth thinking about writing this in Go

yeah, in the next step - we can create the release PRs through bot, and can merge those during the release so that will be handled in Go.

generatebundlefile/hack/update_bundles.sh
jhaanvi5: We can verify it using the git commit of a project in the build tooling repo as an additional check, though I believe latest would always give us whatever the latest was build using the codebuild pipeline.

generatebundlefile/hack/update_bundles.sh
jhaanvi5: Updated.

generatebundlefile/hack/update_bundles.sh
jhaanvi5: Updated, getting latest supported release from [here](https://raw.githubusercontent.com/aws/eks-anywhere-build-tooling/refs/heads/main/release/SUPPORTED_RELEASE_BRANCHES)

generatebundlefile/hack/update_bundles.sh
vivek-koppuru: Yea I think we should add that check just to make sure we are not just doing it from the latest, especially in like situations where a new image was accidentally added with a different version.

generatebundlefile/hack/update_bundles.sh
```diff
@@ -0,0 +1,160 @@
...
+
+  if [[ "$environment" != "staging" && "$environment" != "prod" ]]; then
+    echo "Error: Invalid environment. Use 'staging' or 'prod'" >&2
+    usage
+  fi
+
+  # Check if required tools(aws,jq,yq) are installed
+  check_requirements
+
+  # Generate bundle files for Kubernetes versions 1.27 to 1.31 for given environment
```
vivek-koppuru, line 151: Edit this comment to see where you are reading from in the repo?

generatebundlefile/hack/update_bundles.sh
```diff
@@ -0,0 +1,160 @@
...
+
+  # For cluster autoscaler filter tag based on k8s version (1.28,1.29,etc)
+  if [ "$repo" = "cluster-autoscaler/charts/cluster-autoscaler" ]; then
+    kube_version="1.$k8s_version"
+  # For metrics-server filter tag based on k8s version (1-28,1-29,etc)
+  elif [ "$repo" = "metrics-server/charts/metrics-server" ]; then
+    kube_version="1-$k8s_version"
+  fi
+
+  if [ -n "${kube_version:-}" ]; then
```
vivek-koppuru, line 42: Maybe explain here why your filtering is different in a comment?

generatebundlefile/hack/update_bundles.sh
jhaanvi5: We are also planning to introduce further automation through PR bot, so planning to add this additional check along with that improvement

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #1176, opened 2024-11-20
Removing unwanted feature gate(KubeletCredentialProviders) option

jhaanvi5:
*Description of changes:*
- Removing unwanted feature gate option `KubeletCredentialProviders`. `KubeletCredentialProviders` feature gate was  GA'ed in `1.26` and was removed from `1.28`. For current supported release versions for EKS-A we don't need to add feature-gate option in kubelet-extra-args since the default option is `true` until 1.28 and is not available in later release versions 
Reference: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/
- Updated the unit test

*Testing*
- make test
- Functional testing was done along with this[ PR](https://github.com/aws/eks-anywhere-packages/pull/1177).

### DIFF
--- a/credentialproviderpackage/pkg/configurator/linux/linux.go
+++ b/credentialproviderpackage/pkg/configurator/linux/linux.go
@@ -217,12 +217,6 @@ func (c *linuxOS) createConfig() (string, error) {
 
 func (c *linuxOS) updateKubeletArguments(line string) string {
 	args := ""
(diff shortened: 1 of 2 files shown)


vivek-koppuru:
This PR will address this: https://github.com/aws/eks-anywhere/issues/8984

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #1177, opened 2024-11-22
Add support to update both legacy and default path for kubelet-extra-args for ubuntu

jhaanvi5:
**What did we have before?**

- We used `/etc/sysconfig/kubelet` to update kubelet extra args with credential provider arguments.
- An upstream image-builder update added a new EnvironmentFile (/etc/default/kubelet).
- For Ubuntu, this new file took precedence, causing credential provider arguments to be updated in the wrong place.
- We reverted this change with a [patch](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/kubernetes-sigs/image-builder/patches/0013-Revert-Optionally-source-etc-default-kubelet-in-kube.patch).

**Changes Made:**
Updated the path KUBELET_EXTRA_ARGS file path for Ubuntu to `/etc/default/kubelet` so that credential provider arguments are updated in the right file for Ubuntu, also added KUBELET_EXTRA_ARGS_LEGACY to support older image templates which would still read from `/etc/sysconfig/kubelet`

Once we do packages release, we will also revert the image-builder [patch](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/kubernetes-sigs/image-builder/patches/0013-Revert-Optionally-source-etc-default-kubelet-in-kube.patch) to make it consistent with upstream.

**Testing**
- Created a cluster with Ubuntu image template
- Installed credential provider package using helm 
  - `helm install credential-provider-package oci://public.ecr.aws/personalregistry/credential-provider-package --version 0.4.4-9b0192e048c510b9e3ce36087e992b6d0558a6d4 --set sourceRegistry=public.ecr.aws/personalregistry -n eksa-packages `
- Verified if KUBELET_EXTRA_ARGS are updated in correct path
- Tested for cluster upgrade scenario(k8s version 1.29–1.30)

**Kubelet args in both files before package installation:**
```
root@mgmt-ubuntu-md-0-kczw7-mc289:/home/ec2-user# cat /etc/default/kubelet
KUBELET_EXTRA_ARGS=--pod-infra-container-image=public.ecr.aws/eks-distro/kubernetes/pause:v1.29.8-eks-1-29-22root@mgmt-ubuntu-md-0-kczw7-mc289:/home/ec2-user#

root@mgmt-ubuntu-md-0-kczw7-mc289:/home/ec2-user# cat /etc/sysconfig/kubelet
cat: /etc/sysconfig/kubelet: No such file or directory
```
**Kubelet args in both files after credential provider package was installed:**
```
root@mgmt-ubuntu-md-0-kczw7-mc289:/home/ec2-user# cat /etc/default/kubelet
KUBELET_EXTRA_ARGS=--pod-infra-container-image=public.ecr.aws/eks-distro/kubernetes/pause:v1.29.8-eks-1-29-22 --image-credential-provider-config=/eksa-packages/credential-provider-config.yaml --image-credential-provider-bin-dir=/eksa-packages/

root@mgmt-ubuntu-md-0-kczw7-mc289:/home/ec2-user# cat /etc/sysconfig/kubelet
KUBELET_EXTRA_ARGS= --image-credential-provider-config=/eksa-packages/credential-provider-config.yaml --image-credential-provider-bin-dir=/eksa-packages/
```

### DIFF
--- a/credentialproviderpackage/pkg/configurator/linux/linux.go
+++ b/credentialproviderpackage/pkg/configurator/linux/linux.go
@@ -4,7 +4,6 @@ import (
 	_ "embed"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"strings"
 	"syscall"
@@ -26,7 +25,7 @@ const (
 	basePath              = "/eksa-packages/"
 	credOutFile           = "aws-creds"
 	mountedExtraArgs      = "/node-files/kubelet-extra-args"
-	ubuntuLegacyExtraArgs = "/node-files/kubelet-extra-args-legacy"
+	ubuntuLegacyExtraArgs = "/node-files/ubuntu-legacy-kubelet-extra-args"
 	credProviderFile      = "credential-provider-config.yaml"
 
 	// Binaries
@@ -226,7 +225,7 @@ func (c *linuxOS) createConfig() (string, error) {
 	if err != nil {
 		return "", nil
 	}
-	err = ioutil.WriteFile(dstPath, bytes, 0o600)
+	err = os.WriteFile(dstPath, bytes, 0o600)
 	if err != nil {
 		return "", err
 	}
--- a/credentialproviderpackage/pkg/configurator/linux/linux.go
+++ b/credentialproviderpackage/pkg/configurator/linux/linux.go
@@ -22,11 +22,12 @@ import (
 var credProviderTemplate string
 
 const (
-	binPath          = "/eksa-binaries/"
-	basePath         = "/eksa-packages/"
-	credOutFile      = "aws-creds"
-	mountedExtraArgs = "/node-files/kubelet-extra-args"
-	credProviderFile = "credential-provider-config.yaml"
+	binPath               = "/eksa-binaries/"
+	basePath              = "/eksa-packages/"
+	credOutFile           = "aws-creds"
+	mountedExtraArgs      = "/node-files/kubelet-extra-args"
+	ubuntuLegacyExtraArgs = "/node-files/kubelet-extra-args-legacy"
+	credProviderFile      = "credential-provider-config.yaml"
 
 	// Binaries
 	ecrCredProviderBinary = "ecr-credential-provider"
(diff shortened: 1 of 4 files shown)


credentialproviderpackage/pkg/configurator/linux/linux.go
```diff
@@ -91,10 +93,26 @@ func (c *linuxOS) UpdateCredentialProvider(_ string) error {
 	}
 
 	out := strings.Join(lines, "\n")
-	err = ioutil.WriteFile(c.extraArgsPath, []byte(out), 0o644)
+	err = os.WriteFile(configPath, []byte(out), 0o644)
```
vivek-koppuru, line 96: Just curious why are we changing from ioutil to os right now? Are there other references to ioutil in this file?

credentialproviderpackage/pkg/configurator/linux/linux.go
```diff
@@ -22,31 +22,34 @@ import (
...
-	binPath          = "/eksa-binaries/"
-	basePath         = "/eksa-packages/"
-	credOutFile      = "aws-creds"
-	mountedExtraArgs = "/node-files/kubelet-extra-args"
-	credProviderFile = "credential-provider-config.yaml"
+	binPath               = "/eksa-binaries/"
+	basePath              = "/eksa-packages/"
+	credOutFile           = "aws-creds"
+	mountedExtraArgs      = "/node-files/kubelet-extra-args"
+	ubuntuLegacyExtraArgs = "/node-files/kubelet-extra-args-legacy"
```
vivek-koppuru, line 29: Just to keep naming consistent
```suggestion
	ubuntuLegacyExtraArgs = "/node-files/ubuntu-legacy-kubelet-extra-args"
```

vivek-koppuru (commented):
So these args can be functional in parallel? Also let's create an issue for the followup of removing the patch

credentialproviderpackage/pkg/configurator/linux/linux.go
jhaanvi5: ioutil.readfile was deprecated it basically calls os.readfile, so I updated it, let me know if we should update this with this change. I noticed the same in other files, I can refactor those too.

https://pkg.go.dev/io/ioutil#ReadFile

jhaanvi5:
> So these args can be functional in parallel? Also let's create an issue for the followup of removing the patch

Yes, both of these args will be functional in parallel and created the task for removing the patch: https://github.com/aws/eks-anywhere-internal/issues/2800

vivek-koppuru (approved):

